PR #1215, opened 2014-12-23
GUI-1527 Do not append "(None)" as VPC ID to a newly created security group on th...

kyolee310:
...e Launch Config Wizard

https://eucalyptus.atlassian.net/browse/GUI-1527


kamalgill:
Looks good to merge after UX review passes

PR #1223, opened 2015-01-07
Magic search to replace filter forms and text search bar

dkavanagh:
Here are related issues:
https://eucalyptus.atlassian.net/browse/GUI-1373
https://eucalyptus.atlassian.net/browse/GUI-1472


dkavanagh:
@kamalgill With the last commit, I noticed that the click listener that response when someone clicks outside of the input box is broken. This is probably because the listener is on "search-main-area" and now that isn't a fixed width, but only contains the facets and input.
I also notice things aren't as tight as before. Worse on FF than Chrome. I see facets lower and icons higher and the search-bar is taller than it needs to be with a single row. Try on FF to see it more exaggerated.

kyolee310:
The selenium automated test is failing consistently at Volume and IP tests when trying to look up the search filter, http://10.111.4.12/job/Koala-SUITE/1387/.

The travis CI test's JavsScript Unittest is also failing as well.

kyolee310:
Passed the automated test.

PR #1228, opened 2015-01-09
GUI-1541 don't show addressing mode on launch_more when logged into AWS. enforce public  in code

dkavanagh:
don't show addressing mode on launch_more when logged into AWS. enforce public  in code

https://eucalyptus.atlassian.net/browse/GUI-1541


kamalgill:
Hal, retest this please

PR #1232, opened 2015-01-12
Adding chinese translation and fix the i18n domain for roles.pt and role_view.pt

nathanxu:
https://eucalyptus.atlassian.net/browse/GUI-1546


kyolee310:

kyolee310:
ok to test

PR #1239, opened 2015-01-14
GUI-1547

dkavanagh:
added font-awesome 4.2.0 files

https://eucalyptus.atlassian.net/browse/GUI-1547


kamalgill:
@dkavanagh: We should use Bower to manage the Font Awesome dependency.  /cc @kyolee310

PR #1248, opened 2015-01-23
GUI-1579: Don't pass addressing_type when launching VPC instances on Eucalyptus

kamalgill:
EUCA-10308 modified how the addressing_type param is accepted, rejecting 'private' when a VPC instance is launch.  This fix accounts for the updated behavior.

https://eucalyptus.atlassian.net/browse/GUI-1579

Note: Do not merge this unless GUI-1579 is scoped for 4.1

### DIFF
--- a/eucaconsole/templates/instances/instance_launch.pt
+++ b/eucaconsole/templates/instances/instance_launch.pt
@@ -189,7 +189,7 @@
                                 ${structure:launch_form.monitoring_enabled(**{'ng-model': 'monitoringEnabled'})}
                                 ${launch_form['monitoring_enabled'].label}
                             </div>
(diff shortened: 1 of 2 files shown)


kyolee310:
Additional fix to remove the private addressing option display from the instance launch template.

kyolee310:
@kamalgill  It would be a simple fix to remove the option from the launch more like this page, but I will leave that decision up to you. The current fix has been tested on VPC, non-VPC, and AWS.

kyolee310:
The fix for the launch more like this would require adding

tal:condition="layout.cloud_type == 'euca' and not is_vpc_supported"

to the template, and passing "is_vpc_supported" to class InstanceLaunchMoreView's render_dict on instances.pt

PR #1249, opened 2015-01-23
GUI-1580: Hide the private addressing only option from the launch more like this w...

kyolee310:
...izard for VPC systems
https://eucalyptus.atlassian.net/browse/GUI-1580


kamalgill:
Punted to 4.1.1

PR #1260, opened 2015-01-30
GUI 1536 replace default subpath when it might contain lost path information

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1536

### DIFF
--- a/eucaconsole/views/buckets.py
+++ b/eucaconsole/views/buckets.py
@@ -59,24 +59,26 @@ BUCKET_ITEM_URL_EXPIRES = 300  # Link to item expires in ___ seconds (after page
 BUCKET_NAME_PATTERN = '^[a-z0-9-\.]+$'
 FOLDER_NAME_PATTERN = '^[^\/]+$'
 
+
 class BucketMixin(object):
 
-    def real_path(self, request):
+    @staticmethod
+    def real_path(request):
         if len(request.subpath) == 0:
             return ''
         path = request.environ['PATH_INFO']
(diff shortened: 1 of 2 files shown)


kamalgill:
@dkavanagh: We should add some unit tests for this PR.

eucaconsole/views/buckets.py
```diff
@@ -59,6 +59,24 @@
 BUCKET_NAME_PATTERN = '^[a-z0-9-\.]+$'
 FOLDER_NAME_PATTERN = '^[^\/]+$'
 
+class BucketMixin(object):
+
+    def real_path(self, request):
```
kamalgill: Let's change this to a static method since we're not referencing self in the block, and to simplify adding a unit test for this method.

dkavanagh:
I fixed up these things, but one more issue when uploading to a path that has a double slash. Log says the right thing. Upload goes to path with slashes normalized.

dkavanagh:
in boto, between these 2 lines, a slash has gone missing;
2015-02-02 10:12:48 DEBUG auth_path=/euca-dak/81f67bed-2044-49cf-b848-397a6b5bab92/build//lib/IMG_20140506_084716.jpg?acl
2015-02-02 10:12:48,537 boto [DEBUG]:Method: GET
2015-02-02 10:12:48 DEBUG Method: GET
2015-02-02 10:12:48,537 boto [DEBUG]:Path: /81f67bed-2044-49cf-b848-397a6b5bab92/build/lib/IMG_20140506_084716.jpg?acl

PR #1262, opened 2015-02-02
GUI-1533 Preserve the order of the termination policy options on the scaling group detail page

kyolee310:
https://eucalyptus.atlassian.net/browse/GUI-1533

### DIFF
--- a/eucaconsole/static/js/pages/scalinggroup.js
+++ b/eucaconsole/static/js/pages/scalinggroup.js
@@ -43,7 +43,6 @@ angular.module('ScalingGroupPage', ['AutoScaleTagEditor', 'EucaConsoleUtils'])
         };
         $scope.initController = function (optionsJson) {
             var options = JSON.parse(eucaUnescapeJson(optionsJson));
-            // scalingGroupName, policiesCount
             $scope.scalingGroupName = options.scaling_group_name;
             $scope.policiesCount = options.policies_count;
             $scope.terminationPoliciesOrder = options.termination_policies;
--- a/eucaconsole/static/js/pages/scalinggroup.js
+++ b/eucaconsole/static/js/pages/scalinggroup.js
@@ -220,8 +220,8 @@ angular.module('ScalingGroupPage', ['AutoScaleTagEditor', 'EucaConsoleUtils'])
             });
             // appending duplicated elements into select will sort the items in order
             select.append(newOptions); 
-            // After rearrange the orde of the options, set the values on the select element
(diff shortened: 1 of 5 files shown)


eucaconsole/static/js/pages/scalinggroup.js
```diff
@@ -189,6 +203,44 @@ angular.module('ScalingGroupPage', ['AutoScaleTagEditor', 'EucaConsoleUtils'])
...
+                var mapped = '';
+                angular.forEach(options, function(option) {
+                    if ($(option).val() === policy) {
+                        mapped = option;
+                    } 
+                });
+                return mapped;
+            });
+            // appending duplicated elements into select will sort the items in order
+            select.append(newOptions);
```
kamalgill: @kyolee310: Using jQuery's DOM manipulation methods seems ill-advised here.  There must be a better way.  Have we considered Angular's ngOptions?

kyolee310: The main issue is due to the fact that the chosen widget does not handle the ordering of the selected items, which is a known issue. To get around this issue, the order of the original <select> element's options -- which the chosen is initialized on -- needs to be rearranged at every interaction of the user selection on the chosen widget, and have the chosen widget to be updated after the ordering is done. Since the chosen widget is initialized directly on the <select> element, using Angular calls would only introduce more interference on the operations of the chosen widget.

kamalgill: @kyolee310: Hmm, could we get around the Chosen widget initialization issue by introducing a timeout?  We should avoid using select.append() since it opens up the console to XSS attack vectors.

kyolee310: From previous experience, the Chosen widget works the best when no Angular handlers were involved. Rather than adding timeout calls to get around the Angular conflict issue -- and other possible hacks --, it'd be better to remove the unnecessary components out and keep the implementation under the single component, that is jQuery, which is what the Chosen widget is based on. 

The jQuery call, append(), would only be vulnerable if it takes a new string or input form the page, however, in this case, it is only doing the reordering of the existing HTML option elements -- if there were XXS code within such options, it would be triggered beforehand, regardless of the append() call.

kamalgill: @kyolee310: Yeah, I see your point.  Since we control the termination policies list in our code, we could let the .append() slide here.  I'm still curious if the ngOptions approach would have worked, since that would be potentially cleaner and safer.

PR #1273, opened 2015-02-09
GUI-1590 handle smart-refresh better with multiple getItems() callers

dkavanagh:
handle smart-refresh better with multiple getItems() callers

https://eucalyptus.atlassian.net/browse/GUI-1590

The changes are these:
 - add state variable to indicate smart refresh is already happening, so we can avoid starting a new one.
 - reset that variable when no state requiring smart refresh is in the table
 - pass optional value into getItems() so that we can know when it was called from existing smart refresh timeout

This works well. It refreshes properly with no filters. As filters are added, the single refresh timeout is the only one running and the new filters are used. If the transient item is filtered out, smart refresh halts after the next cycle. If a transient item is brought back in by a filter change, smart refresh initiates properly again. This change fixes the mult-fetch problem seen prior.


dkavanagh:
hal, retest this please

PR #1276, opened 2015-02-09
GUI-1590 handle smart-refresh better with multiple getItems() callers

dkavanagh:
handle smart-refresh better with multiple getItems() callers

https://eucalyptus.atlassian.net/browse/GUI-1590

The changes are these:

add state variable to indicate smart refresh is already happening, so we can avoid starting a new one.
reset that variable when no state requiring smart refresh is in the table
pass optional value into getItems() so that we can know when it was called from existing smart refresh timeout
This works well. It refreshes properly with no filters. As filters are added, the single refresh timeout is the only one running and the new filters are used. If the transient item is filtered out, smart refresh halts after the next cycle. If a transient item is brought back in by a filter change, smart refresh initiates properly again. This change fixes the mult-fetch problem seen prior.

### DIFF
--- a/eucaconsole/static/js/pages/landingpage.js
+++ b/eucaconsole/static/js/pages/landingpage.js
@@ -10,6 +10,7 @@ angular.module('LandingPage', ['CustomFilters', 'ngSanitize', 'MagicSearch'])
         $http.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
         $scope.items = [];
         $scope.itemsLoading = true;
+        $scope.runningSmartRefresh = false;
         $scope.unfilteredItems = [];
         $scope.filterKeys = [];
         $scope.sortBy = '';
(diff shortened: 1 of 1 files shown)


dkavanagh:
hal, retest this please

eucaconsole/static/js/pages/landingpage.js
```diff
@@ -157,7 +158,13 @@ angular.module('LandingPage', ['CustomFilters', 'ngSanitize', 'MagicSearch'])
                 });
                 // Auto-refresh items if any are in a transitional state
                 if ($scope.transitionalRefresh && transitionalCount > 0) {
-                    $timeout(function() { $scope.getItems(); }, 5000);  // Poll every 5 seconds
+                    if ($scope.runningSmartRefresh == false || okToRefresh !== undefined) {
```
kamalgill: The runningSmartRefresh check can be simplified (i.e. no need for the "== false" part.

For the second check, instead of testing if okToRefresh is undefined, why not cast okToRefresh as a boolean (defaulting to false) by adding the following at the top of the function block...  `okToRefresh = okToRefresh || false`.  

Or better yet, allow the getItems to accept an 'options' object, then test for options.okToRefresh, assuming you've added `options = options || {};` at the top of the function block.  This approach is more future-proof since it's trivial to pass additional options down the road, with no change required in the function signature.

dkavanagh: Sure, I was trying to be explicit as we are in the python. (on the runningSmartRefresh check)

I think putting in framework that may never be used has been the cause of bloat in past projects I've been involved in. I don't think wrapping a single value in a associative array buys us much considering this method has remained "argument free" this long.
About the test. It would be great if javascript had parameter typing and/or default values. I'm not in favor of adding more code to make up for that necessarily. We have a simple check in place right now. I'd rather that be a check for false, but undefined is the only default allowed by the language, so I think that's pretty clear to javascript developers. This happens all the time for testing select options.

PR #1277, opened 2015-02-09
Added storeTextIfPresent operation and a condition to delete-volume-on-detail-page test for it to pass if deleted volume is removed from landing page

euca-qa-gui-tester:


kyolee310:

PR #1278, opened 2015-02-09
Fixes for GUI-1559 and other unicode-related tickets

kamalgill:
Fixes for the following tickets:
https://eucalyptus.atlassian.net/browse/GUI-1559
https://eucalyptus.atlassian.net/browse/GUI-1566
https://eucalyptus.atlassian.net/browse/GUI-1567
https://eucalyptus.atlassian.net/browse/GUI-1568
https://eucalyptus.atlassian.net/browse/GUI-1569
https://eucalyptus.atlassian.net/browse/GUI-1570
https://eucalyptus.atlassian.net/browse/GUI-1571
https://eucalyptus.atlassian.net/browse/GUI-1595

### DIFF
--- a/eucaconsole/caches.py
+++ b/eucaconsole/caches.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Copyright 2014 Eucalyptus Systems, Inc.
 #
 # Redistribution and use of this software in source and binary forms,
@@ -26,8 +27,8 @@
 
 import inspect
 import pylibmc
-from hashlib import sha1
 from dogpile.cache import make_region
+
 
 def euca_key_generator(namespace, fn):
     if fn is None:
@@ -35,6 +36,7 @@ def euca_key_generator(namespace, fn):
     else:
         args = inspect.getargspec(fn)
         has_self = args[0] and args[0][0] in ('self', 'cls')
+
     def generate_key(*arg):
         # generate a key:
         # "namespace_arg1_arg2_arg3..."
@@ -44,10 +46,11 @@ def euca_key_generator(namespace, fn):
 
         # return cache key
         # apply sha1 to obfuscate key contents
-        #return sha1(key).hexdigest()
+        # return sha1(key).hexdigest()
         return key
 
     return generate_key
+
 
 def invalidate_cache(cache, namespace, *arg):
     key = euca_key_generator(namespace, None)(*arg)
--- a/eucaconsole/views/keypairs.py
+++ b/eucaconsole/views/keypairs.py
@@ -157,7 +157,7 @@ class KeyPairView(BaseView):
     def keypair_create(self):
         if self.keypair_form.validate():
             name = self.request.params.get('name')
-            # WORKAROUND: Boto seems to barf on names with non-ascii chars in conn.create_key_pair(name)
+            # NOTE: Normalize non-ascii chars, which aren't allowed in key pairs
             name = self.normalize_unicode(name)
             location = self.request.route_path('keypair_view', id=name)
             with boto_error_handler(self.request, location):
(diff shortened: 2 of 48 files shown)


dkavanagh:
hal, retest this please

eucaconsole/caches.py
```diff
@@ -26,15 +27,16 @@
 
 import inspect
 import pylibmc
-from hashlib import sha1
```
dkavanagh: Looking at this, I'd actually prefer to add the key obfuscation back. I remember backing it out to make it simpler to debug some cache key issues, though I still think it's a good idea, especially with a shared memcached install.

dkavanagh:
trying to review the changes, wading through so much code-prettification that it makes it hard to see what's fixed for this bug. I recommend doing that on a separate branch (and it's a good idea if we ever want to add flake8 to travis-ci)

eucaconsole/views/keypairs.py
```diff
@@ -157,9 +157,11 @@ def get_keypair_names(self):
     def keypair_create(self):
         if self.keypair_form.validate():
             name = self.request.params.get('name')
+            # NOTE: Normalize non-ascii chars, which aren't allowed in key pairs
+            name = self.normalize_unicode(name)
```
dkavanagh: What does it mean to do this? If we should not allow unicode characters, let's not.

dkavanagh:
The work for the i18n stuff looks good. I question the need to normalize inputs that should be filtered already.

eucaconsole/views/keypairs.py
kamalgill: @dkavanagh: The pattern-based filtering doesn't seem to filter out unicode chars.  For example, I was able to add a key pair named 'keypair-ÅÅÅ" which threw a 500 error before the normalize_unicode() was in place.

eucaconsole/caches.py
kamalgill: @dkavanagh: Understood, feel free to add it back when you reimplement it.  I was just removing an unused import here.

kamalgill:
Hal, retest this please

eucaconsole/caches.py
dkavanagh: I was hoping you could not remove the commented out code below (and hopefully the import) so that I could write a separate ticket and get that code enabled for another release (4.2?) more simply

eucaconsole/views/keypairs.py
dkavanagh: If that's the case, I think adjusting the filter would be appropriate. From what I read, that filter should limit to characters in the range specified.

eucaconsole/caches.py
kamalgill: @dkavanagh: I didn't remove any commented code in this module

dkavanagh: oic. sorry. I remembered seeing the change in it and was mostly thinking about with the import removed, I'd have to remember where sha1 came from again.

eucaconsole/views/keypairs.py
kamalgill: @dkavanagh: Where do you see the pattern `{'pattern': '^[\x20-\x7F]{1,255}$'}`?  Looking at keypair_view.pt, the pattern is `{'pattern': '^[^\/\\\]{1,255}$'}`, which does complicate things since we'll need to modify that regex to accept any ASCII char except slashes.

dkavanagh: well dang, I was looking in master. That has the fix I put in and I forgot this was for maint-4.1. I think think we should adjust the regex so that the user will be told if they enter an invalid character. For 4.1.1, perhaps we need to accept that range minus the slashes?

eucaconsole/caches.py
dkavanagh: ^((?![\x2F])[\x20-\x7F]){1,255}$

eucaconsole/views/keypairs.py
dkavanagh: ^((?![\x2F])[\x20-\x7F]){1,255}$

kamalgill:
Hal, retest this please

eucaconsole/views/securitygroups.py
```diff
@@ -270,6 +270,8 @@ def securitygroup_delete(self):
     def securitygroup_create(self):
         if self.securitygroup_form.validate():
             name = self.request.params.get('name')
+            # NOTE: Normalize non-ascii chars, which aren't allowed in security groups
+            name = self.normalize_unicode(name)
```
dkavanagh: If we're filtering in javascript w/ regex, this is un-necessary. Let's remove and the method in the base view.

PR #1279, opened 2015-02-10
GUI-1602 Fixed the misuse of ternary operator (?:) on security group rules JS wid...

kyolee310:
...get

https://eucalyptus.atlassian.net/browse/GUI-1602

### DIFF
--- a/eucaconsole/static/js/jasmine-spec/spec_security_group_rules.js
+++ b/eucaconsole/static/js/jasmine-spec/spec_security_group_rules.js
@@ -315,4 +315,132 @@ describe("SecurityGroupRules", function() {
         });
     });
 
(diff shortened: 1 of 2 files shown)


kamalgill:
@kyolee310: Is GUI-1602 scoped for 4.1.1?

kyolee310:
The JIRA ticket is still in 'unconfirmed' state, but this fix should be merged immediately since it addresses the apparent mistake in the usage of the ternary operator. It was a lucky case that the faulty logic was generating the correct functional behavior.

The fix has been tested on Chrome and Firefox, and also the unit-test for the ternary operations have been included to verify the expected output for various conditions.

euca-qa-gui-tester:

PR #1363, opened 2015-04-23
Create test-PR-monitor

euca-qa-gui-tester:


kamalgill:
@aliceh: We can close this PR since we've tracked down the root cause of the Jenkins failures (see https://github.com/eucalyptus/eucaconsole/pull/1366 for the cause and https://github.com/eucalyptus/eucaconsole/pull/1381 for the fix)

PR #1367, opened 2015-04-24
Gui 1476 refactor launch wizards based on what users use most often

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1476


eucaconsole/templates/instances/instance_launch.pt
```diff
@@ -81,6 +81,22 @@
                             <hr />
                             ${panel('tag_editor', show_name_tag=False, leftcol_width=3, rightcol_width=9)}
                             <hr />
+                            <div class="row controls-wrapper readonly">
+                                <div class="columns">
+                                    <h6 i18n:translate="">User data</h6>
+                                </div>
+                                <div class="small-12 columns value">
+                                    <input type="radio" id="inputtype" name="inputtype" value="text" ng-model="inputtype"/>
+                                    <label id='userdatalabel' i18n:translate="">Enter text</label>
```
kamalgill: @dkavanagh: This page has two labels with the same id attribute.  Note that valid HTML requires ids to be unique per document.  Also, please use double quotes for attributes to be consistent, and consider adding a "for" attribute to the label elements.  See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label for details.

eucaconsole/templates/instances/instance_launch_more.pt
```diff
@@ -100,6 +100,25 @@
...
+                    <div class="row controls-wrapper readonly">
+                        <div class="small-4 columns">
+                            <label>
+                                <span i18n:translate="">User data</span>
+                                <i class="helptext-icon" i18n:translate="title" title="User data file may not exceed 16 KB" data-tooltip=""></i>
+                            </label>
+                        </div>
+                        <div class="small-8 columns value">
+                            <input type="radio" id="inputtype" name="inputtype" value="text" ng-model="inputtype"/>
+                            <label id='userdatalabel' i18n:translate="">Enter text</label>
```
kamalgill: @dkavanagh My comments w.r.t. the label elements in instance_launch.pt also apply here.

eucaconsole/templates/launchconfigs/launchconfig_wizard.pt
```diff
@@ -75,6 +75,22 @@
                                 </div>
                             </div>
                             <hr />
+                            <div class="row controls-wrapper readonly">
+                                <div class="columns">
+                                    <h6 i18n:translate="">User data</h6>
+                                </div>
+                                <div class="small-12 columns value">
+                                    <input type="radio" id="inputtype" name="inputtype" value="text" ng-model="inputtype"/>
+                                    <label id='userdatalabel' i18n:translate="">Enter text</label>
```
kamalgill: @dkavanagh My comments w.r.t. the label elements in instance_launch.pt also apply here.  Since this text block is repeated in numerous places, perhaps we should create a panel or macro to keep things DRY?

PR #1370, opened 2015-04-28
GUI-1611: CloudWatch Charts - Instance Monitoring page

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-1611

Tested on...
- Safari 8, Chrome 42, and Firefox 37 on Mac OS 10.10
- IE 11 on Windows 7 (64-bit)


kamalgill:
Hal, retest this please

kamalgill:
Jenkins failure is the random waitForVisible failure in the create-volume-from-snapshot-on-snapshot-lp test, which should be safe to ignore.

Makefile
```diff
@@ -13,3 +13,8 @@ translations:
 # Handy way to test if our Chameleon-based XHTML templates are valid XML (for i18n)
 xmltest:
 	python setup.py extract_messages
+
+ci:
```
dkavanagh: Curious why not adding to Gruntfile.js instead?

eucaconsole/static/js/pages/cloudwatch_charts.js
```diff
@@ -0,0 +1,215 @@
...
+    vm.largeChartDuration = 3600;
+    vm.largeChartGranularity = 300;
+    vm.metricTitleMapping = {};
+    vm.chartsList = [];
+    vm.granularityChoices = [];
+    vm.originalDurationGranularitiesMapping = {};
+    vm.durationGranularitiesMapping = {};
+    vm.largeChartMetric = '';
+    vm.largeChartLoading = false;
+    vm.initController = initController;
```
dkavanagh: I didn't realize we had to link all of the functions to the vm. That's not pretty, but I suppose shorter than $scope. What's vm stand for anyway?

kamalgill: @dkavanagh: https://github.com/johnpapa/angular-styleguide#controlleras-with-vm

eucaconsole/static/sass/pages/cloudwatch_charts.scss
```diff
@@ -0,0 +1,73 @@
...
+@import "../includes/eucavariables";
+
+.cloudwatch-charts {
+    .toprow {
+        @include flexbox-mixin;
+        .blurb {
+            @include flexgrow-mixin(1);
+        }
+        .actions {
+            display: inline-block;
```
dkavanagh: I think you should have flexgrow-mixin(0) here, right? Is that just un-necessary, or more complete?

kamalgill: @dkavanagh: flexgrow doesn't seem to be necessary here, so let's leave it out for now.

eucaconsole/templates/instances/instance_monitoring.pt
```diff
@@ -0,0 +1,104 @@
...
+        <form method="post" id="monitoring-form" action="${request.route_path('instance_monitoring_update', id=instance.id)}">
+            ${structure:monitoring_form['csrf_token']}
+        </form>
+        <!--! Modal dialog for large chart -->
+        ${panel('cloudwatch_chart_dialog', duration_choices=duration_choices, statistic_choices=statistic_choices)}
+    </div>
+</div>
+
+<div metal:fill-slot="tail_js">
+    <script src="${request.static_path('eucaconsole:static/js/thirdparty/jquery/chosen.jquery.min.js')}"></script>
```
dkavanagh: Is this actually used?

kamalgill: @dkavanagh: Looks as though it isn't, so I'll remove this line.  Thanks for catching that!

eucaconsole/static/sass/pages/cloudwatch_charts.scss
dkavanagh: There are plenty of things that "seem to" work, but I wonder if there's a spec'ed behavior for an item in flexbox that does not define flex-grow (when others do)? If it's spec-ed to default to 0, fine, but otherwise, we should ensure that by specifying.

kamalgill: @dkavanagh: I'd refer to the flexbox spec for how the flexgrow behavior works ;-)

kamalgill:
Hal, retest this please

Makefile
kamalgill: @dkavanagh : We'd need to add a dependency such as grunt-shell to run `python setup.py test`, so I opted for the simpler approach via make.

eucaconsole/views/cloudwatchapi.py
```diff
@@ -0,0 +1,178 @@
...
+                x=time.mktime(dt_object.timetuple()) * 1000,  # Milliseconds since Unix epoch
+                y=amount
+            ))
+
+        # Sort by timestamp to avoid chart anomalies
+        json_stats = sorted(json_stats, key=itemgetter('x'))
+
+        return dict(
+            unit=unit,
+            results=[dict(key=metric, values=json_stats)],
```
dkavanagh: Out of curiosity, why wrap a single dict in an array?

kamalgill: @dkavanagh: Good question--that's the format expected by NVD3.  Each item in the list is a line series, but since there's only one per chart the response data structure may seem a bit verbose.  We will probably need to refactor this when we need support multiple line series per chart, but the wireframes haven't spec'd that case for 4.2 so this is good enough for now.

PR #1371, opened 2015-04-28
GUI-1539 on AWS, display only the same network elastic IPs for selections on the associate IP dialog

kyolee310:
https://eucalyptus.atlassian.net/browse/GUI-1539

### DIFF
--- a/eucaconsole/static/js/pages/instances.js
+++ b/eucaconsole/static/js/pages/instances.js
@@ -28,7 +28,7 @@ angular.module('InstancesPage', ['LandingPage', 'EucaConsoleUtils'])
             var options = JSON.parse(eucaUnescapeJson(optionsJson));
             if (options.hasOwnProperty('addresses_json_items_endpoint')) {
                 $scope.addressesEndpoint = options.addresses_json_items_endpoint;
-            }; 
+            } 
             $scope.getIPAddresses(); 
             $scope.initChosenSelectors();
             $('#file').on('change', $scope.getPassword);
--- a/eucaconsole/static/js/pages/instances.js
+++ b/eucaconsole/static/js/pages/instances.js
@@ -5,11 +5,14 @@
  */
 
 angular.module('InstancesPage', ['LandingPage', 'EucaConsoleUtils'])
-    .controller('InstancesCtrl', function ($scope, $http, eucaHandleError) {
+    .controller('InstancesCtrl', function ($scope, $timeout, $http, eucaUnescapeJson, eucaHandleError) {
         $scope.instanceID = '';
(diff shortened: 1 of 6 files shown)


eucaconsole/views/instances.py
```diff
@@ -326,7 +332,9 @@ def instances_disassociate_ip_address(self):
                 msg = _(u'Successfully disassociated the IP from the instance.')
                 self.request.session.flash(msg, queue=Notification.SUCCESS)
             return HTTPFound(location=self.location)
-        return self.render_dict
+        msg = _(u'Failed to disassociate the IP to the instance.')
```
kamalgill: @kyolee310: The message should probably be `Failed to disassociate the IP address from the instance`, but please confirm with @JMoLo

eucaconsole/static/js/pages/instances.js
```diff
@@ -52,6 +60,13 @@ angular.module('InstancesPage', ['LandingPage', 'EucaConsoleUtils'])
             $scope.publicDNS = instance.public_dns_name;
             $scope.platform = instance.platform;
             $scope.ipAddress = instance.ip_address;
+            if (action === 'associate-ip-to') {
+                $scope.adjustIPAddressOptions(instance);
```
kamalgill: @kyolee310: If we only need to call the adjustIPAddressOptions() method on AWS, we should wrap that with an AWS check, right?

kyolee310: The function does not only filter the elastic IPs by the network type, but also removes the elastic IPs that are already associated with instances.

eucaconsole/views/instances.py
kyolee310: @JMoLo let us know what the error message should be here.

JMoLo: The one @kamalgill suggested is fine

kyolee310: The error messages have been updated.

kyolee310:
retest please

kyolee310:
@aliceh could you restart this test please?

aliceh:
retest this please

aliceh:
@kyolee310 the ip-operations sequence failed at the same place as the previous one. Namely, when associate IP dialog is called from instance detail page then the drop-down is clicked, the IP to be associated  is not found on the list. I tested the same with console from master and it worked.

kamalgill:
Jenkins failure is the problematic `disassociate-ip-from-instance-on-ip-lp` test, which has a random failure with the notifications placeholder.  Safe to ignore, so merging this PR...

PR #1373, opened 2015-05-04
GUI-1702: Remove unnecessary ng-app attributes

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-1702


kyolee310:
retest this please.

kyolee310:
The test failure on the snapshot landing page seems like the test issue itself. When tested manually, it was confirmed that the volume was able to be created from the snapshot landing page option. Good to merge.

PR #1374, opened 2015-05-05
GUI-1718, GUI-1719: Display appropriate monitoring status and tab title on AWS when basic or detailed monitoring is enabled

kamalgill:
Fixes: https://eucalyptus.atlassian.net/browse/GUI-1718
Implements: https://eucalyptus.atlassian.net/browse/GUI-1719


eucaconsole/models/auth.py
```diff
@@ -213,7 +214,8 @@ def _euca_connection(_clchost, _port, _access_id, _secret_key, _token, _conn_typ
                 conn.auth_region_name = 'Eucalyptus'
 
             setattr(conn, 'APIVersion', api_version)
-            conn.https_validate_certificates = validate_certs
+            if conn:
```
dkavanagh: Looking at this in context, I can't see why we'd test in this case, but not for the other code that uses conn. The mission of this method is to create a connection, so it should always be there, right?

eucaconsole/models/auth.py
```diff
@@ -135,7 +135,8 @@ def _aws_connection(_region, _access_key, _secret_key, _token, _conn_type):
                     _region, aws_access_key_id=_access_key, aws_secret_access_key=_secret_key, security_token=_token)
             elif conn_type == 'iam':
                 return None
-            conn.https_validate_certificates = validate_certs
+            if conn:
```
dkavanagh: conn will never be None (the case for IAM returns directly)

kamalgill: @dkavanagh: The condition was added for the InstanceMonitoringView unit tests, where conn is None.

dkavanagh: We should probably have a discussion about what kinds of changes to our production code are acceptable to facilitate unit tests. I wonder if there is another way around this? If we probably shouldn't mask errant behavior. If conn was None at that point in the code (in production), I'd want to know about it (as close to the error as possible).

kamalgill: @dkavanagh: Sorry, I should have been more clear.  Note that conn is None in the unit test, which is perfectly fine since there's no need to build a conn object in the unit test.  Given that, it's safe to add the condition here IMO.  We probably won't need to add these conditions when more of the moto-based unit tests are added, and we have a separate epic to track those tests.

PR #1379, opened 2015-05-06
GUI-1728: Ensure form field help text/icons are implemented consistently across the app

kamalgill:
Also set 16K char maxlength in userdata textareas.

Fixes https://eucalyptus.atlassian.net/browse/GUI-1728

### DIFF
--- a/eucaconsole/templates/instances/instance_launch.pt
+++ b/eucaconsole/templates/instances/instance_launch.pt
@@ -87,13 +87,17 @@
                                 </div>
                                 <div class="small-12 columns value">
                                     <input type="radio" id="inputtype" name="inputtype" value="text" ng-model="inputtype"/>
-                                    <label class='userdatalabel' i18n:translate="">Enter text</label>
-                                    <textarea id="userdata" name="userdata" ng-model="userData" ng-show="inputtype=='text'"></textarea>
+                                    <label class="userdatalabel" i18n:translate="">Enter text</label>
+                                    <textarea id="userdata" name="userdata"
+                                              ng-model="userData" ng-show="inputtype=='text'" maxlength="16000"></textarea>
                                     <br/>
                                     <input type="radio" id="inputtype" name="inputtype" value="file" ng-model="inputtype"/>
-                                    <label class='userdatalabel' i18n:translate="">Upload file</label>
+                                    <label class="userdatalabel">
+                                        <span i18n:translate="">Upload file</span>
+                                        <span tal:condition="getattr(launch_form.userdata_file, 'help_text', None)" class="helptext-icon"
+                                            data-tooltip="" title="${launch_form.userdata_file.help_text}"></span>
+                                    </label>
                                     <input id="userdata_file" type="file" ng-model="userDataFile" ng-disabled="isNotStopped" name="userdata_file" disabled="disabled" ng-show="inputtype=='file'"/><span id="userdatatype"></span>
-                                    
                                 </div>
                             </div>
                             <hr />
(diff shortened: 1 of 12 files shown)


eucaconsole/templates/instances/instance_launch.pt
```diff
@@ -87,13 +87,17 @@
                                 </div>
                                 <div class="small-12 columns value">
                                     <input type="radio" id="inputtype" name="inputtype" value="text" ng-model="inputtype"/>
-                                    <label class='userdatalabel' i18n:translate="">Enter text</label>
-                                    <textarea id="userdata" name="userdata" ng-model="userData" ng-show="inputtype=='text'"></textarea>
+                                    <label class="userdatalabel" i18n:translate="">Enter text</label>
+                                    <textarea id="userdata" name="userdata"
+                                              ng-model="userData" ng-show="inputtype=='text'" maxlength="16000"></textarea>
```
dkavanagh: 16K is actually 16,384, but with base64 encoding, I think the practical limit ends up being a bit lower. Not sure this maxlength will really avoid errors from the back end in cases where the encoded data is longer.

eucaconsole/templates/panels/form_field_row.pt
```diff
@@ -5,15 +5,15 @@
         <div class="large-${leftcol_width_large} small-${leftcol_width} columns" tal:condition="leftcol_width">
             <label class="${'right' if inline else 'left'}">
                 ${field.label.text}&nbsp;<span tal:condition="field.flags.required" class="req">*</span>
-                <span tal:condition="getattr(field, 'label_help_text', None)" class="helptext-icon"
-                    data-tooltip="" title="${field.label_help_text}">?</span>
+                <span tal:condition="getattr(field, 'help_text', None)" class="helptext-icon"
+                    data-tooltip="" title="${field.help_text}"></span>
```
dkavanagh: I would use the nested icon element here and a few lines down, much like what was done around line 33 (for consistency)

eucaconsole/templates/instances/instance_launch.pt
kamalgill: @dkavanagh: I'm open to suggestions on how to handle this.

dkavanagh:
might want to merge master in here to get tests passing as well.

eucaconsole/templates/instances/instance_launch.pt
dkavanagh: one way would be don't check on the client, or do, but use a much higher number (like the worst case base64 encoded 16k string). Or, write some javascript to validate by base64 encoding and check length on that, but that might be "above and beyond"

eucaconsole/templates/instances/instance_launch_more.pt
```diff
@@ -103,18 +103,19 @@
...
-                                <i class="helptext-icon" i18n:translate="title" title="User data file may not exceed 16 KB" data-tooltip=""></i>
-                            </label>
+                            <label i18n:translate="">User data</label>
                         </div>
                         <div class="large-10 small-8 columns value">
                             <input type="radio" id="inputtype" name="inputtype" value="text" ng-model="inputtype"/>
-                            <label class='userdatalabel' i18n:translate="">Enter text</label>
-                            <textarea id="userdata" name="userdata" ng-show="inputtype=='text'"></textarea>
+                            <label class="userdatalabel" i18n:translate="">Enter text</label>
+                            <textarea id="userdata" name="userdata" ng-show="inputtype=='text'" maxlength="16384"></textarea>
```
dkavanagh: I think the point was missed from my previous comments. The user will not be able to enter user data this long and get it to work. Base64 encoding expansion prevents it, so this limit is not terribly useful.

PR #1380, opened 2015-05-07
GUI-1729: Removed custom_filters.js script tags in the templates

kyolee310:
https://eucalyptus.atlassian.net/browse/GUI-1729


kyolee310:
The automated test failed on 'create-volume-from-snapshot-on-snapshot-lp', which is a known issue in current test.

PR #1381, opened 2015-05-08
GUI-1735: Convert Availability Zone selector from Chosen widget to standard drop-down on create volume page

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-1735

Fixes recurring Jenkins failure on create volume page's zone selector

### DIFF
--- a/eucaconsole/static/sass/pages/volume.scss
+++ b/eucaconsole/static/sass/pages/volume.scss
@@ -26,7 +26,7 @@ $volume-status-color-attaching: lighten($euca-darkblue, 20%);
 // Volume page
 input#size { width: 4rem; }
 input#name { width: 50%; }
-select#zone { width: 200px; }
+select#zone { width: 50%; }
 
 
 // Volume status
--- a/eucaconsole/static/sass/pages/volume.scss
+++ b/eucaconsole/static/sass/pages/volume.scss
@@ -24,6 +24,10 @@ $volume-status-color-attaching: lighten($euca-darkblue, 20%);
 
 // ***********
 // Volume page
+input#size { width: 4rem; }
+input#name { width: 50%; }
+select#zone { width: 200px; }
+
 
 // Volume status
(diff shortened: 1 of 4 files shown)


eucaconsole/static/sass/pages/volume.scss
```diff
@@ -24,6 +24,10 @@ $volume-status-color-attaching: lighten($euca-darkblue, 20%);
 
 // ***********
 // Volume page
+input#size { width: 4rem; }
+input#name { width: 50%; }
+select#zone { width: 200px; }
```
dkavanagh: curious, but why is 1 rem, 1 percent and 1 px?

kamalgill: @dkavanagh: no particular reason.  I only changed the zone value from a percentage to 200px to better handle the switch to a full-page width.  200px is a somewhat arbitrary value for the zone input, but it works well across our small, medium and large breakpoints.

dkavanagh: Has @JMoLo looked at this? I guess I'm less bothered by the different unit (though it makes it hard to get consistent output), but looking at the form has 4 different size fields and bringing name/snapshot/zone to match would visually be nicer.

kamalgill: @dkavanagh: @JMoLo hasn't looked at it.  We can reassign this to her if she needs to look at it, but I only changed the width of the zone selector here since we moved away from a Chosen widget.  I don't quite understand why there's a concern using different CSS units for those different fields.  Perhaps we can discuss next week?

JMoLo: I would definitely prefer that the margins of inputs line up whenever possible. Now that we are not having to code so frantically as we did in 4.0, I am going to start paying more attention to the visual design and general usability of all our pages.

PR #1383, opened 2015-05-11
GUI 1548 create launch config like this (launch config)

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1548


eucaconsole/templates/launchconfigs/launchconfig_create_more.pt
```diff
@@ -0,0 +1,122 @@
...
+                <form action="${request.route_path('launchconfig_create')}" method="post"
+                      data-abide="abide" enctype="multipart/form-data" id="launchconfig-more-form"
+                      tal:define="html_attrs_val {'pattern': '^[^\/\\\]{1,255}$'};">
+                    ${structure:create_form['csrf_token']}
+                    ${structure:create_form['image_id']}
+                    <input type="hidden" name="from_url" value="${request.route_path('launchconfig_more', id=launchconfig_name)}"/>
+                    ${panel('form_field', field=create_form['name'], leftcol_width=3, rightcol_width=8, autofocus='', ng_attrs={'model': 'launchconfigName'}, **html_attrs_val)}
+                    <div class="row controls-wrapper readonly">
+                        <div class="large-2 small-4 columns"><label i18n:translate="">Image name</label></div>
+                        <div class="large-10 small-8 columns value">${image.name}</div>
```
kamalgill: @dkavanagh: Need an ng-non-bindable attr here

eucaconsole/templates/launchconfigs/launchconfig_create_more.pt
```diff
@@ -0,0 +1,122 @@
...
+                        ${panel('form_field', field=create_form['associate_public_ip_address'], leftcol_width=4, rightcol_width=8)}
+                    </div>
+                    ${panel('form_field', field=create_form['keypair'], leftcol_width=4, rightcol_width=8)}
+                    ${panel('form_field', field=create_form['securitygroup'], leftcol_width=4, rightcol_width=8)}
+                    ${panel('form_field', field=create_form['role'], leftcol_width=4, rightcol_width=8)}
+                    <div class="row controls-wrapper readonly">
+                        <div class="large-2 small-4 columns">
+                            <label>
+                                <span i18n:translate="">User data</span>
+                                <i class="helptext-icon" i18n:translate="title" title="User data file may not exceed 16 KB" data-tooltip=""></i>
```
kamalgill: @dkavanagh: The i18n attribute should be `i18n:attributes="title"`

eucaconsole/templates/launchconfigs/launchconfig_create_more.pt
```diff
@@ -0,0 +1,122 @@
...
+                            </label>
+                        </div>
+                        <div class="large-10 small-8 columns value">
+                            <input type="radio" id="inputtype" name="inputtype" value="text" ng-model="inputtype"/>
+                            <label class='userdatalabel' i18n:translate="">Enter text</label>
```
kamalgill: @dkavanagh: Let's use double quotes for attributes here and elsewhere to be consistent

eucaconsole/templates/launchconfigs/launchconfig_create_more.pt
```diff
@@ -0,0 +1,122 @@
...
+                            </div>
+                            <hr/>
+                            ${panel('bdmapping_editor', image=image, snapshot_choices=snapshot_choices)}
+                        </div>
+                    </div>
+                    <div class="row">
+                        <div class="large-2 small-4 columns">&nbsp;</div>
+                        <div class="large-10 small-8 columns field inline">
+                            <button type="submit" class="button" id="save-changes-btn">
+                                <span i18n:translate="">Create Launch Configuration</span><span ng-show="instanceNumber > 1">s</span>
```
kamalgill: We may need to escape the '>' in `instanceNumber > 1` to avoid breaking i18n

eucaconsole/templates/panels/form_field_row.pt
```diff
@@ -5,8 +5,9 @@
         <div class="large-${leftcol_width_large} small-${leftcol_width} columns" tal:condition="leftcol_width">
             <label class="${'right' if inline else 'left'}">
                 ${field.label.text}&nbsp;<span tal:condition="field.flags.required" class="req">*</span>
-                <span tal:condition="getattr(field, 'label_help_text', None)" class="helptext-icon"
-                    data-tooltip="" title="${field.label_help_text}">?</span>
+                <span tal:condition="getattr(field, 'label_help_text', None)" data-tooltip="" title="${field.label_help_text}">
```
kamalgill: @dkavanagh: Heads up that GUI-1728 has set the help text attribute as 'help_text' everywhere, so we should probably merge https://github.com/eucalyptus/eucaconsole/pull/1379 , merge master into this branch, and make the changes here before we merge this PR.

eucaconsole/templates/launchconfigs/launchconfig_create_more.pt
dkavanagh: actually, I can remove instanceNumber entirely.. that's left over from "launch more like this".

eucaconsole/templates/panels/form_field_row.pt
kamalgill: @dkavanagh: On second thought, let's back out these changes to avoid merge conflicts with PR 1379 since that PR fixes the form field help icons.

eucaconsole/views/launchconfigs.py
```diff
@@ -436,12 +437,47 @@ def launchconfig_new(self):
         self.render_dict['preset'] = self.request.params.get('preset')
         return self.render_dict
 
+    @view_config(route_name='launchconfig_more', renderer=TEMPLATE_MORE, request_method='GET')
+    def launchconfig_more(self):
+        name = self.request.matchdict.get('id')
+        autoscale_conn = self.get_connection(conn_type='autoscale')
+        with boto_error_handler(self.request):
+            lc = autoscale_conn.get_all_launch_configurations(names=[name])
+            self.launch_config = lc[0]
```
kamalgill: @dkavanagh: It's bad practice to set an instance attribute outside an init method.  BTW, PyCharm would have made this blatantly obvious.  Let's move the create more launch configs code to a distinct class/view rather than shoving it into the CreateLaunchConfigView class.

kamalgill: @dkavanagh: We should still move the launch more to a dedicated view to keep those code paths distinct to better avoid regressions.

eucaconsole/templates/launchconfigs/launchconfig_create_more.pt
dkavanagh: All this time, I thought if you named an attribute in "i18n:translate", that it translated that. Did that change, or did I miss something? I've been doing it this way for the past 2 years!

kamalgill: @dkavanagh: Nothing has changed.  `i18n:translate` is for marking text that is the content of an element, while i18n:attributes is to mark text in an attribute value.  And looking at accounts.pt lines 37-53 as an example, you've marked those strings just fine elsewhere.  See http://chameleon.readthedocs.org/en/latest/reference.html#id49

eucaconsole/views/launchconfigs.py
dkavanagh: I don't see your point about regressions. I mean, this could easily be in another view, but nothing in this call would affect the rest of the code in this class. Plus the template served by this ends up calling into the same view to create the launch config, so it's packaged in the right place, IMHO.

eucaconsole/templates/launchconfigs/launchconfig_create_more.pt
kamalgill: @dkavanagh: I've opened https://eucalyptus.atlassian.net/browse/GUI-1737 to track the i18n issues.  Will work on a fix shortly.

PR #1387, opened 2015-05-14
GUI -1753: Fixed ids of cloudformation and stacks menuitems in the nav menu

aliceh:
https://eucalyptus.atlassian.net/browse/GUI-1753


kamalgill:
Hal, retest this please

PR #1401, opened 2015-05-28
Gui 1722 support dns enabled clouds plus other fixes for hp discover demo

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1722

Supports automatic discovery of dns-enablement by trying auth using "tokens." prefix on host first and failing back to plain "clchost" in config.
other fixes include:
 - enable delete dialog on stack view
 - set up proper wrapping for stack events and resources tables
 - remove xmlns attr for instance monitoring tab
 - add compatibilities param when creating stack


kamalgill:
Hal, retest this please

PR #1406, opened 2015-05-29
GUI-1483, GUI-1620, GUI-1648: ELB detail pages

kamalgill:
Implements:
- https://eucalyptus.atlassian.net/browse/GUI-1483
- https://eucalyptus.atlassian.net/browse/GUI-1620
- https://eucalyptus.atlassian.net/browse/GUI-1648


eucaconsole/static/js/pages/elb.js
```diff
@@ -1,51 +1,78 @@
...
         $scope.initController = function (optionsJson) {
             var options = JSON.parse(eucaUnescapeJson(optionsJson));
-            $scope.hasImage = options.has_image;
+            $scope.setInitialValues(options);
             $scope.setWatch();
-            $scope.setFocus();
-            if (!$scope.hasImage) {
-                $('#image-missing-modal').foundation('reveal', 'open');
+            $timeout(function(){
+                $scope.isInitComplete = true;
```
dkavanagh: Is this needed? It's only used in listeners. Was this to prevent some listeners from firing early?

kamalgill: @dkavanagh: Kyo added that code to prevent Foundation validation form raising false negative warnings, so we should probably leave it in to be safe.

dkavanagh: I just think it's not obvious what it's for, so warrants at least a 1 line comment.

dkavanagh:
hal, retest this please

PR #1407, opened 2015-06-01
Gui 1667 Converted to split buttons were needed

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1667

 - added split buttons to bucket, bucket_content and keypair pages.
 - converted paste button to icon button
 - added needed repository field in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "eucaconsole",
   "version": "4.1.0",
+  "repository": "https://github.com/eucalyptus/eucaconsole",
   "devDependencies": {
     "grunt": "~0.4.5",
     "grunt-cli": "~0.1.13",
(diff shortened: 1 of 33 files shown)


package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "eucaconsole",
   "version": "4.1.0",
```
kamalgill: @dkavanagh: I understand this isn't relevant to the ticket, but should we set the version here to 4.2.0?

dkavanagh: I think we'd typically do that closer to release, though this isn't used anywhere outside this file since we don't publish to bower.io

dkavanagh:
hal, retest this please

kamalgill:
Hal, retest this please

PR #1417, opened 2015-06-05
GUI-1821 added additional check in case instance profile has no role(s)

dkavanagh:
added additional check in case instance profile has no role(s)

https://eucalyptus.atlassian.net/browse/GUI-1821


aliceh:
@dkavanagh  whoops, just saw that there is no image installed on the test setup. apologies, installing now.

PR #1418, opened 2015-06-05
GUI 1637 Added support for AWS template import.

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1637
https://eucalyptus.atlassian.net/browse/GUI-1771


tests/stacks/test_aws_convert.py
```diff
@@ -0,0 +1,460 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import unittest
+import simplejson as json
+
+class TestAWSConvert(unittest.TestCase):
+    def test_aws_convert(self):
+        from eucaconsole.views.stacks import StackWizardView
+        # this is an actual aws template that contains RDS resources
+        template = """
```
kamalgill: @dkavanagh: We should probably move this template to an external file and have the unit test read from that file

kamalgill:
Hal, retest this please

eucaconsole/views/stacks.py
```diff
@@ -29,20 +29,23 @@
 
 """
 import simplejson as json
+from simplejson import JSONDecodeError
```
gholms: This exception is not available in the version of simplejson that ships with RHEL 6.

aliceh:
Good to merge. Random SeleniumRC glitch.

kamalgill:
This PR is also for https://eucalyptus.atlassian.net/browse/GUI-1649

kamalgill:
Hal, retest this please

dkavanagh:
hal, retest this please

eucaconsole/static/js/pages/stack_wizard.js
```diff
@@ -137,6 +154,28 @@ angular.module('StackWizard', ['TagEditor', 'EucaConsoleUtils', 'localytics.dire
                 $scope.checkRequiredInput();
             });
             $scope.$watch('inputtype', function(){
```
kamalgill: @dkavanagh: No need to change it here, but the function can be spec'd  as `function(newValue, oldValue)` , where newValue is $scope.inputtype (so the switch statement would be `switch (newValue) {... }`.  
See the example at https://docs.angularjs.org/api/ng/type/$rootScope.Scope#example

kamalgill:
The Jenkins failures have nothing to do with the code changes here, so this should be safe to merge.

PR #1423, opened 2015-06-10
GUI-1828: Prevent unexpected logout on Create ELB wizard when IAM policy prevents listing of certificates

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-1828

### DIFF
--- a/eucaconsole/forms/elbs.py
+++ b/eucaconsole/forms/elbs.py
@@ -464,27 +464,33 @@ class CertificateForm(BaseSecureForm):
     certificates_error_msg = _(u'Certificate is required')
     certificates = wtforms.SelectField(
         label=_(u'Certificate name'),
-        validators=[validators.InputRequired(message=certificates_error_msg)],
     )
 
-    def __init__(self, request, conn=None, iam_conn=None, elb_conn=None, **kwargs):
+    def __init__(self, request, conn=None, iam_conn=None, elb_conn=None, can_list_certificates=True, **kwargs):
         super(CertificateForm, self).__init__(request, **kwargs)
         self.conn = conn
         self.iam_conn = iam_conn
         self.elb_conn = elb_conn
+        self.can_list_certificates = can_list_certificates
         self.set_error_messages()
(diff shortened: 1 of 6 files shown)


eucaconsole/forms/elbs.py
```diff
@@ -464,27 +464,33 @@ class CertificateForm(BaseSecureForm):
     certificates_error_msg = _(u'Certificate is required')
     certificates = wtforms.SelectField(
         label=_(u'Certificate name'),
-        validators=[validators.InputRequired(message=certificates_error_msg)],
```
dkavanagh: Would cert still be required if cert form is validated? Wouldn't we just not validate this form if no cert required? We only require a cert for SSL termination, AFAIK

eucaconsole/templates/dialogs/select_certificate_dialog.pt
```diff
@@ -73,20 +73,28 @@
...
             <p>
                 <span i18n:translate="">
-                    SSL certificates are required for secure(HTTPS or SSL) listeners.  Choose or upload a pem-encoded certificate below.
+                    SSL certificates are required for secure(HTTPS or SSL) listeners.
+                </span>
+                <span i18n:translate="" tal:condition="can_list_certificates">
+                    Choose or upload a pem-encoded certificate below.
+                </span>
+                <span i18n:translate="" tal:condition="not can_list_certificates">
+                    Upload a pem-encoded certificate below.
```
dkavanagh: Since all certs must be uploaded with IAM, if they can't list certificates, I doubt they can upload one.

eucaconsole/forms/elbs.py
kamalgill: @dkavanagh: The certificate form only displays for HTTPS/SSL protocols, and the validators are conditionally added in the set_certificate_choices() method below, so I think we're fine here.

dkavanagh:
That failure on storeText for span.euca.username-label wouldn't be bad if it didn't bypass a bunch of test code.

kamalgill:
Hal, retest this please

PR #1424, opened 2015-06-11
GUI-1832: Avoid wiping out original port when switching protocols on the ELB detail page Health Checks tab

kamalgill:
Potential fix for https://eucalyptus.atlassian.net/browse/GUI-1832


kamalgill:
Leaving this branch open since there's more work to do here.

PR #1438, opened 2015-06-29
Gui 1587 Added options to view shared bucket, download/upload from/to shared bucket.

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1587


dkavanagh:
Looks like the failures were not critical and likely random.

aliceh:
Good to merge. One random fail in set_all_to_list view.

aliceh:
Good to merge.

eucaconsole/templates/buckets/buckets.pt
```diff
@@ -20,9 +20,26 @@
...
+                    </li>
+                    <li>
+                        <a id="upload-object-btn" i18n:translate="" target="_blank"
+                            href="${request.route_path('bucket_upload', name='__shared__', subpath='')}">Upload Object(s)</a>
+                    </li>
+                </ul>
             </div>
             <div metal:fill-slot="tile_header">
-                <a ng-href="{{ item.contents_url }}" class="has-bucket-icon">
+                <a ng-href="{{ item.contents_url }}" class="has-shared-icon" ng-show="{{ item.shared }}">
```
kamalgill: @dkavanagh: Unlike ng-href, ng-show and ng-hide expect an expression as the attribute value, so we should remove the curly braces here and elsewhere.  See https://docs.angularjs.org/api/ng/directive/ngShow

dkavanagh: good catch.

dkavanagh:
fixed curly braces in 4 places. Good to merge?

PR #1440, opened 2015-06-30
GUI-1856 fixed 2 places that needed additional checks with CLI users deletes role without removing instance profile

dkavanagh:
fixed 2 places that needed additional checks with CLI users deletes role without removing instance profile

https://eucalyptus.atlassian.net/browse/GUI-1856


aliceh:
Good to merge. 3 of the sequences failed "set all to list view" test due to random Selenium RC failure.

PR #1454, opened 2015-07-07
GUI-1841 fix delete object from top level of bucket

dkavanagh:
only add prefix + / if prefix is not empty string

https://eucalyptus.atlassian.net/browse/GUI-1841


aliceh:
Good to merge.

PR #1456, opened 2015-07-07
GUI-1146 added meta refresh tag to reload page every 29 minutes

dkavanagh:
added meta refresh tag to reload page every 29 minutes

https://eucalyptus.atlassian.net/browse/GUI-1146

### DIFF
--- a/eucaconsole/templates/login.pt
+++ b/eucaconsole/templates/login.pt
@@ -1,7 +1,7 @@
 <metal:block use-macro="main_template" i18n:domain="eucaconsole">
 
 <div metal:fill-slot="head_css" tal:omit-tag="">
-    <meta http-equiv="refresh" content="1740">
+    <meta http-equiv="refresh" content="${login_refresh}">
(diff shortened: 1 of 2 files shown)


kamalgill:
@dkavanagh: Instead of hard-coding the refresh to every 29 minutes, we should set the refresh interval based on the `session.timeout` value from console.ini since it's configurable.  Perhaps set the refresh interval to `session.timeout - 60` seconds?

dkavanagh:
I'll have to test to verify the session.timeout is what's affecting the csrf token expiration first.

kamalgill:
@dkavanagh: The `session.timeout` value specifies the idle timeout duration, which is what GUI-1146 covers AFAICT.

dkavanagh:
Well, we know the idle timeout is non-functional since the app will log you out during usage. I'm not making any assumptions without verifying.

dkavanagh:
I was able to verify that the session.timeout is directly related to how often the login becomes non-functional, so I'll use a value that's 60 seconds less than session.timeout.

PR #1463, opened 2015-07-09
GUI-1860 improved name regex and made validation error more accurate

dkavanagh:
improved name regex and made validation error more accurate

https://eucalyptus.atlassian.net/browse/GUI-1860


eucaconsole/forms/stacks.py
```diff
@@ -39,8 +39,8 @@ class StacksCreateForm(BaseSecureForm):
     """Stacks creation form.
        Only need to initialize as a secure form to generate CSRF token
     """
-    name_error_msg = _(u"""Name is required and may contain lowercase letters,
-        numbers, and/or hyphens and should not be longer than 255 characters""")
+    name_error_msg = _(u"""Name is required, must begin with a letter, and may contain only letters,
+        numbers, and/or hyphens. Length must not exceed than 255 characters.""")
```
kamalgill: @JMoLo, @dkavanagh : Should this be "Length must not exceed 255 characters." here?

JMoLo: Ugh, yes. Can you fix it please, kamal? I was concentrating so hard on the rest of the string that I missed that. Thanks!

On Jul 14, 2015, at 5:10 PM, Kamal Gill <notifications@github.com<mailto:notifications@github.com>> wrote:


In eucaconsole/forms/stacks.py<https://github.com/eucalyptus/eucaconsole/pull/1463#discussion_r34634538>:

> @@ -39,8 +39,8 @@ class StacksCreateForm(BaseSecureForm):
>      """Stacks creation form.
>         Only need to initialize as a secure form to generate CSRF token
>      """
> -    name_error_msg = _(u"""Name is required and may contain lowercase letters,
> -        numbers, and/or hyphens and should not be longer than 255 characters""")
> +    name_error_msg = _(u"""Name is required, must begin with a letter, and may contain only letters,
> +        numbers, and/or hyphens. Length must not exceed than 255 characters.""")


@JMoLo<https://github.com/JMoLo>, @dkavanagh<https://github.com/dkavanagh> : Should this be "Length must not exceed 255 characters." here?

�

PR #1472, opened 2015-07-12
GUI-1706, GUI-1800: Fix whitespace/wrapping issues on login page

kamalgill:
Fixes whitespace issues and wrapping of tab titles on login page.

https://eucalyptus.atlassian.net/browse/GUI-1706
https://eucalyptus.atlassian.net/browse/GUI-1800

### DIFF
--- a/eucaconsole/templates/login.pt
+++ b/eucaconsole/templates/login.pt
@@ -47,12 +47,14 @@
                 <dl class="tabs" id="login-tabs" data-tab="">
                     <dd id="euca-tab" class="active">
                         <a href="#eucalyptus">
-                            <span class="hide-for-small" i18n:translate="">Log in to</span>&nbsp;<span i18n:translate="">Eucalyptus</span>
+                            <span class="hide-for-small" i18n:translate="">Log in to Eucalyptus</span>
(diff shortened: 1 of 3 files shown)


eucaconsole/templates/login.pt
```diff
@@ -109,7 +110,7 @@
                             <input type="hidden" id="duration" name="duration" value="${duration}" />
                             <div class="row controls-wrapper controls_${field.name}" tal:repeat="field aws_login_form">
                                 <div class="small-12 columns field">
-                                    <label for="top-label" class="left">${field.label.text}<span class="req">*</span></label>
+                                    <label for="top-label" class="left stack-label">${field.label.text}<span class="req">*</span></label>
```
dkavanagh: What is the for attribute doing in this case? I know it was there before, but unless it links to the input field, I'm not sure the use of it.

kamalgill: @dkavanagh : Good question. The `for="top-label"` was added in commit https://github.com/eucalyptus/eucaconsole/commit/6408b6c6ccb104601d7badadab3c26648d9de898 (and prior to that it was for="right-label", which isn't accurate either).  Good catch -- will fix.

aliceh:
retest this, please

PR #1473, opened 2015-07-13
GUI-1805, GUI-1822: ELB Detail Page - Instance selector improvements

kamalgill:
Implements:
- https://eucalyptus.atlassian.net/browse/GUI-1805
- https://eucalyptus.atlassian.net/browse/GUI-1822

### DIFF
--- a/eucaconsole/static/sass/widgets/instance_selector.scss
+++ b/eucaconsole/static/sass/widgets/instance_selector.scss
@@ -25,3 +25,7 @@ $instance-status-color-terminating: lightgrey;
 // ELB instance status
 .status.InService {background-color: $instance-status-color-running; }
 .status.OutOfService { background-color: $instance-status-color-stopped; }
+
+
+// Other styles
+.fi-alert { color: darkred; }
--- a/eucaconsole/static/css/pages/elb.css
+++ b/eucaconsole/static/css/pages/elb.css
@@ -150,7 +150,7 @@ a.nolink:hover { color: #5a5a5a !important; cursor: inherit; }
 
 p.error { color: darkred; }
 
-table.instance-selector-table { width: 100%; }
+table.table { width: 100%; }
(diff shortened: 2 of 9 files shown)


eucaconsole/static/sass/widgets/instance_selector.scss
```diff
@@ -25,3 +25,7 @@ $instance-status-color-terminating: lightgrey;
 // ELB instance status
 .status.InService {background-color: $instance-status-color-running; }
 .status.OutOfService { background-color: $instance-status-color-stopped; }
+
+
+// Other styles
+.fi-alert { color: darkred; }
```
dkavanagh: I'm a little surprised this isn't a more widely used adjustment.

kamalgill: @dkavanagh: yeah, we should move this to eucaconsole.scss.  I'll open a new ticket for that.

dkavanagh: Is this the alert icon we should be using? https://github.com/eucalyptus/eucaconsole/blob/master/eucaconsole/static/sass/euca-font-awesome.scss#L71
Granted, that was for an alert dialog but I would think icon source and color should be consistent, even if size is different.

kamalgill: @dkavanagh: I wasn't aware of us switching the alert icon to Font Awesome, since we've been using the Foundation alert icon everywhere else.  Could you add your notes to https://eucalyptus.atlassian.net/browse/GUI-1888 and we can discuss with @JMoLo how we want to handle the alert icons?

JMoLo: we've used both, whichever way y'all want is OK with me, the look is not substantially different between the two.

PR #1474, opened 2015-07-14
GUI-1885 improve image param checking in aws templates

dkavanagh:
ignore image params when FindInMap is found and referenced map contains any emi- reference. Also don't perform aws check on sample templates

https://eucalyptus.atlassian.net/browse/GUI-1885


dkavanagh:
Merging this to more major stack fixes into master. We can revisit the PR later if issues are uncovered.

PR #1480, opened 2015-07-16
GUI-1721 added xhr call to create bucket if new one specified. If name conflicts, display warning

dkavanagh:
added xhr call to create bucket if new one specified. If name conflicts, display warning

https://eucalyptus.atlassian.net/browse/GUI-1721

### DIFF
--- a/eucaconsole/static/js/pages/instance_create_image.js
+++ b/eucaconsole/static/js/pages/instance_create_image.js
@@ -6,7 +6,8 @@
 
 // Create Image page includes the Tag Editor 
 angular.module('InstanceCreateImage', ['TagEditor'])
-    .controller('InstanceCreateImageCtrl', function ($scope, $timeout) {
+    .controller('InstanceCreateImageCtrl', function ($scope, $http, $timeout) {
+        $http.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
         $scope.form = $('#create-image-form');
         $scope.expanded = false;
         $scope.name = '';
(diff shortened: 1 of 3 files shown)


dkavanagh:
hal, retest this please

eucaconsole/static/js/pages/instance_create_image.js
```diff
@@ -70,7 +72,19 @@ angular.module('InstanceCreateImage', ['TagEditor'])
                 $scope.checkRequiredInput();
             });
             $scope.$watch('s3_bucket', function () {
+                $("#bucket-exists-error").css('display', 'none');
                 $scope.checkRequiredInput();
+                if ($scope.s3_bucket.length > 0) {
+                    var data = "csrf_token=" + $('#csrf_token').val() + "&bucket_name=" + $scope.s3_bucket;
+                    $http({method:'POST', url:$scope.bucketCreateUrl, data:data,
```
kamalgill: @dkavanagh: Ugh, issuing an HTTPS post on every keystroke is a huge red flag.  See how I handled this in https://github.com/eucalyptus/eucaconsole/commit/f302c6c6961ab31aba1a73dc87de7744c1857629 (which hasn't been merged yet but hopefully will land in 4.2).

dkavanagh: Except it doesn't do it on every keystroke. This is not a text input box, it's chosen. Only when the user select an item, or enters a new value (completely)

kamalgill: @dkavanagh: Issuing the create bucket API call on drop-down selection or custom text input entry is also ill-advised, especially without an explicit submit button click.  Let's discuss this with @JMoLo at our next meeting, since there are security implications here.

JMoLo: @kamalgill and I discussed this for another ticket, and decided to do this there as an explicit link to a modal box like we do to create key pairs in the launch wizards, for example. It's probably more work to do it that way, but if it's more secure, and solves other issues, that's a good way to go.

kamalgill:
closing w/o merging due to security/implementation concerns

PR #1482, opened 2015-07-16
GUI-1549 added ufshost to image cache key

dkavanagh:
added ufshost to image cache key

https://eucalyptus.atlassian.net/browse/GUI-1549

### DIFF
--- a/eucaconsole/views/__init__.py
+++ b/eucaconsole/views/__init__.py
@@ -197,7 +197,7 @@ class BaseView(object):
         )
 
     @long_term.cache_on_arguments(namespace='images')
-    def _get_images_cached_(self, _owners, _executors, _ec2_region, acct):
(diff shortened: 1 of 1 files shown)


eucaconsole/views/__init__.py
```diff
@@ -236,8 +236,9 @@ def get_images(self, conn, owners, executors, ec2_region):
             acct = self.request.session.get('access_id', '')
         if 'amazon' in owners or 'aws-marketplace' in owners:
             acct = ''
+        ufshost = self.get_connection().host
```
kamalgill: @dkavanagh : We only need to add the ufshost to the cache key for Eucalyptus, so this line should probably be `ufshost = self.get_connection().host if self.cloud_type == 'euca' else ''`.

kamalgill:
Hal, retest this please

ccassler:
Hal, retest this please

PR #1484, opened 2015-07-16
GUI-1894: Update instance type default for launch wizards on AWS…

kamalgill:
…since m1.small is no longer a choice there.

Fixes https://eucalyptus.atlassian.net/browse/GUI-1894


kamalgill:
Closing w/o merging.  In light of https://eucalyptus.atlassian.net/browse/GUI-1898 , we'll need to restore the prior-generation instance types, which will fix GUI-1894 as well.

PR #1488, opened 2015-07-17
GUI-1698 added another unit test and some fixes to tested code

dkavanagh:
added another unit test and some fixes to tested code

https://eucalyptus.atlassian.net/browse/GUI-1698


kamalgill:
Nice work @dkavanagh!  And thanks for moving the test template to an external file!

PR #1491, opened 2015-07-23
GUI-1596 install nginx and start via console init process

mbacchi:
@gholms and @kamalgill

Please review, I've incorporated Garrett's recommendations and it works well as far as I have seen.


euca-qa-gui-tester:

euca-qa-gui-tester:

dkavanagh:
The changes here look right assuming nginx will be installed by package scripts. I wonder if the change to console.default.ini is better handled by "sed" in the install script which also configures nginx?

mbacchi:
Thanks for the comments and review @dkavanagh.  You are right about considering the "python setup.py install" case when making changes like this for RPM based installs.

In an earlier version of this change I had used sed to modify the nginx.conf but @gholms recommended against doing so in a %post RPM script.  I would appreciate everyone's input so we can agree on a compromise that would work best for both environments.

mbacchi:
Vic's suggested I patch the console.ini during RPM build vs. shipping it modified for both source and rpm packages.  The rpmbuild command will now patch it in the build process and what gets shipped in the RPM will show "session.secure = true", but the source distribution will still show "session.secure = false".  @gholms, @dkavanagh, @kamalgill please let me know what you think.

kamalgill:
@mbacchi: sounds good to me

euca-qa-gui-tester:

dkavanagh:
I'm good with this.

aliceh:
test this please

PR #1498, opened 2015-07-29
GUI-1900 fixed stack delete link on secondary tabs

dkavanagh:
fixed stack delete link on secondary tabs

https://eucalyptus.atlassian.net/browse/GUI-1900
https://eucalyptus.atlassian.net/browse/GUI-1901

### DIFF
--- a/eucaconsole/templates/stacks/stack_view.pt
+++ b/eucaconsole/templates/stacks/stack_view.pt
@@ -156,7 +156,7 @@
                         </span>
                         <ul id="actions-dropdown2" class="f-dropdown" data-dropdown-content="">
                             <li>
-                                <a id="delete-stack-action" data-reveal-id="delete-stack-modal"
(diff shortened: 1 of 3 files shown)


kamalgill:
Hal, retest this please

eucaconsole/templates/stacks/stack_view.pt
```diff
@@ -187,9 +188,8 @@
                         </span>
                         <ul id="actions-dropdown3" class="f-dropdown" data-dropdown-content="">
                             <li>
-                                <a id="delete-user-action" i18n:translate="">
-                                    Delete stack
-                                </a>
+                                <a id="delete-stack-action" data-reveal-id="delete-stack-modal"
```
kamalgill: @dkavanagh: An id attribute should be unique for an HTML document, but we have three instances of `id="delete-stack-action"`, which may lead to JS bugs and/or selenium testing issues.  Could we append the tab label to the id attrs here?

PR #1500, opened 2015-07-30
Gui 1625 improve expired and change password error reporting and redirect

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1625
https://eucalyptus.atlassian.net/browse/GUI-1670


eucaconsole/views/managecredentials.py
```diff
@@ -51,7 +51,7 @@ class ManageCredentialsView(BaseView, PermissionCheckMixin):
     def __init__(self, request):
         super(ManageCredentialsView, self).__init__(request)
         self.changepassword_form = EucaChangePasswordForm(self.request)
-        referrer = urlparse(self.request.url).path
+        referrer = urlparse(self.request.headers['REFERER']).path
```
kamalgill: @dkavanagh: Did you test this on a console fronted by a reverse proxy  (e.g. nginx)?  I don't think 'REFERER' would work there, since the proper referring URL would be provided by the 'X_FORWARDED_FOR' header.

eucaconsole/views/managecredentials.py
```diff
@@ -127,17 +133,12 @@ def handle_changepassword(self):
                     # the logging here and below is really very useful when debugging login problems.
                     logging.info("http error "+str(vars(err)))
                     if err.msg == u'Unauthorized':
-                        self.changepassword_form_errors.append(u'Invalid user/account name and/or password.')
+                        msg = u'Invalid user/account name and/or password.'
```
kamalgill: @dkavanagh: We should mark this string for i18n

eucaconsole/views/managecredentials.py
```diff
@@ -127,17 +133,12 @@ def handle_changepassword(self):
...
-                        self.changepassword_form_errors.append(u'No response from host ' + host)
-        return dict(
-            changepassword_form=changepassword_form,
-            changepassword_form_errors=self.changepassword_form_errors,
-            password_expired=True if self.request.params.get('expired') == 'true' else False,
-            came_from=self.came_from,
-            account=account,
-            username=username
-        )
+                        msg = u'No response from host ' + host
```
kamalgill: @dkavanagh: We should mark this string for i18n

eucaconsole/views/managecredentials.py
```diff
@@ -88,14 +88,19 @@ def handle_changepassword(self):
...
         duration = self.request.registry.settings.get('session.cookie_expires')
-        account = "huh?"
-        username = "what?"
+        account = self.request.params.get('account')
+        username = self.request.params.get('username')
         auth = self.get_euca_authenticator()
         changepassword_form = EucaChangePasswordForm(self.request, formdata=self.request.params)
+        self.location = "{0}?{1}&{2}&{3}&{4}".format(
+            self.request.route_path('managecredentials'),
+            'expired=' + self.request.params.get('expired').lower(),
```
kamalgill: @dkavanagh: Is 'expired' guaranteed to be in request.params?  If it isn't, the lower() method will return a 500 error.

dkavanagh: Oddly, it doesn't throw a 500 when expired isn't there, but I think removing "lower()" is safe anyway.

eucaconsole/views/managecredentials.py
dkavanagh: I'm firing up an instance to test this on. The old code clearly didn't work and REFERER did, but I'll see what's needed behind nginx and if that works w/ elb as well.

dkavanagh: I set this up w/ nginx and when hitting that on port 80 (on the same host though), this still works.

kamalgill: @dkavanagh: Does REFERER work when nginx is on a different host?

dkavanagh: I just set this up w/ nginx on 1 host and the console on a 2nd host. I was on an object detail page, selected "manage credentials" and changed my password. I was sent back to the same page. I logged out and back in, went to elastic ips page and did the same. it worked!

kamalgill: @dkavanagh: Good to know, thanks for confirming!

PR #1502, opened 2015-07-31
GUI-1917 removed SSL from this template

dkavanagh:
removed SSL from this template

https://eucalyptus.atlassian.net/browse/GUI-1917


dkavanagh:
hal, retest this please

PR #1503, opened 2015-07-31
GUI-1904 added protocol whitelist for loading templates from aribrary url

dkavanagh:
added protocol whitelist for loading templates from aribrary url
also improved error reporting if can't read from URL

https://eucalyptus.atlassian.net/browse/GUI-1904

### DIFF
--- a/eucaconsole/views/stacks.py
+++ b/eucaconsole/views/stacks.py
@@ -632,7 +632,10 @@ class StackWizardView(BaseView):
                 template_body = files[0].file.read()
                 template_name = files[0].name
             elif template_url:  # read from url
-                protocol = template_url[:template_url.find('://')]
+                idx =  template_url.find('://')
+                if idx == -1:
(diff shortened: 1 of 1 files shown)


eucaconsole/views/stacks.py
```diff
@@ -626,7 +632,13 @@ def parse_store_template(self):
                 template_body = files[0].file.read()
                 template_name = files[0].name
             elif template_url:  # read from url
-                template_body = urllib2.urlopen(template_url).read()
+                protocol = template_url[:template_url.find('://')]
```
kamalgill: @dkavanagh: We should probably use urlparse here -- see https://docs.python.org/2/library/urlparse.html

dkavanagh: I did consider that. Since I only care about the protocol, I opted for this more direct approach.

dkavanagh:
already included in PR1504

PR #1506, opened 2015-08-04
Gui 1903 added stack-id prefix to templates after creation. added collision detection and retry for stack bucket

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1903

### DIFF
--- a/eucaconsole/templates/stacks/stack_view.pt
+++ b/eucaconsole/templates/stacks/stack_view.pt
@@ -166,8 +166,7 @@
                         <div class="row controls-wrapper readonly" tal:condition="template_name">
                             <div class="small-2 columns"><label i18n:translate="" class="right">Name</label></div>
                             <div class="small-10 columns value">
-                                <a href="${request.route_path('bucket_item_details', name=template_bucket, subpath=template_key)}"
-                                   i18n:translate="">${template_name}</a>
+                                <a href="${request.route_path('bucket_item_details', name=template_bucket, subpath=template_key)}">${template_name}</a>
                             </div>
                         </div>
                         <div class="row controls-wrapper readonly">
--- a/eucaconsole/templates/stacks/stack_view.pt
+++ b/eucaconsole/templates/stacks/stack_view.pt
@@ -163,6 +163,13 @@
                     <div class="columns section">
                         &nbsp;
                         <h6 i18n:translate="">Template</h6>
+                        <div class="row controls-wrapper readonly" tal:condition="template_name">
+                            <div class="small-2 columns"><label i18n:translate="" class="right">Name</label></div>
+                            <div class="small-10 columns value">
+                                <a href="${request.route_path('bucket_item_details', name=template_bucket, subpath=template_key)}"
+                                   i18n:translate="">${template_name}</a>
+                            </div>
+                        </div>
                         <div class="row controls-wrapper readonly">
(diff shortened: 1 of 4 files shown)


kamalgill:
Hal, retest this please

eucaconsole/templates/stacks/stack_view.pt
```diff
@@ -163,6 +163,13 @@
                     <div class="columns section">
                         &nbsp;
                         <h6 i18n:translate="">Template</h6>
+                        <div class="row controls-wrapper readonly" tal:condition="template_name">
+                            <div class="small-2 columns"><label i18n:translate="" class="right">Name</label></div>
+                            <div class="small-10 columns value">
+                                <a href="${request.route_path('bucket_item_details', name=template_bucket, subpath=template_key)}"
+                                   i18n:translate="">${template_name}</a>
```
kamalgill: @dkavanagh: We should remove the i18n:translate attr here

eucaconsole/templates/stacks/stack_wizard.pt
```diff
@@ -61,6 +61,13 @@
                             <h6 i18n:translate="">Template</h6>
                             <div class="row">
                                 <div class="small-12 columns value">
+                                    <p class="description" i18n:translate="">
+                                        Your Cloudformation template will be stored in an S3 bucket.
```
kamalgill: @dkavanagh: Shouldn't this be "Your CloudFormation template..."?

eucaconsole/views/stacks.py
```diff
@@ -55,6 +57,37 @@
 TEMPLATE_BODY_LIMIT = 460800
 
 
+class StackMixin(object):
+    def get_create_template_bucket(self, create=True):
```
kamalgill: @dkavanagh: We should default to create=False IMO

eucaconsole/views/stacks.py
```diff
@@ -55,6 +57,37 @@
...
+                if create:
+                    bucket = s3_conn.create_bucket(bucket_name)
+                else:
+                    bucket = s3_conn.get_bucket(bucket_name)
+                return bucket
+            except BotoServerError as err:
+                if err.code != 'BucketAlreadyExists' and err.code != 'AccessDenied':
+                    raise err
+        raise JSONError(status=500, message=_(
+            u'Cannot create S3 bucket to store your Cloudformation template due to namespace collision. '
```
kamalgill: @dkavanagh: 'Cloudformation' should be 'CloudFormation', right?

dkavanagh: https://eucalyptus.atlassian.net/wiki/display/3UD/Create+stack?src=contextnavpagetreemode

following the case there but I think you're right: http://aws.amazon.com/documentation/cloudformation/
@JMoLo

eucaconsole/views/stacks.py
dkavanagh: I understand. I just went with the common use case for the default though.

kamalgill: @dkavanagh: The two places where you're invoking this pass create=False, so that seems to be the common use case and really should be the default since we don't want to mistakenly create a bucket via a GET request again.

dkavanagh: I get it. There are 5 places it's called using the default and that's why I made True the default

PR #1511, opened 2015-08-06
GUI-1899: Link generation and sharing improvements for S3 objects

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-1899


eucaconsole/views/buckets.py
```diff
@@ -921,6 +935,27 @@ def get_bucket_item(self):
             item = self.bucket.get_key(u'{0}/'.format(item_key_name))
         return item
 
+    def get_unsigned_url(self):
+        """Returns unsigned URL for object with query string removed"""
+        item_url = self.bucket_item.generate_url(expires_in=0, query_auth=False)
```
dkavanagh: fwiw, using default arguments is just fine here. A maintainer in the future would ask what these do w/o comments.

PR #1516, opened 2015-08-13
DOC-1560: created LB CREATE topics.

tvo318:
Please add hooks for these LB topics. Thanks!


dkavanagh:
The help content for the elb wizard is a single page, so unfortunately, these need to be combined. @JMoLo , we ought to look at how we present help in the wizards, maybe after 4.2

tvo318:
If we combine all 4 pages of the create and all 4 pages of the details, they will be 4-5 pages each. Do you really want that, @JMoLo ?

dkavanagh:
I'm not convinced combining them is the best either. Since help is given space under the summary area in the wizards, it already looks pretty squished and long.

tvo318:
Help is at the bottom for the details pages, which makes users scroll way down to find what they are looking for.

JMoLo:
Can you display the associated help for whichever tab the user is currently on, rather than combining into a single file? These are really long. 

We need to optimize to wide screens, too, so at the super big breakpoint these can go back up to the left side. That's not a 4.2 thing, though, more like 4.3.

tvo318:
So, I combine the files for now? and split them out again for 4.3?

JMoLo:
I'd say leave them separated for now and let the guys figure out how to display them.

tvo318:
ok, thanks, Jenny. I'll keep my eyes and ears out if I need to do anything to accommodate.

tvo318:
The hooks for the help pages look really good. Thank you all!

PR #1518, opened 2015-08-21
GUI 1622 support for ELB access logs

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1622


eucaconsole/static/js/pages/elb_wizard.js
```diff
@@ -295,6 +302,24 @@ angular.module('BaseELBWizard').controller('ELBWizardCtrl', function ($scope, $h
...
+        $scope.$watch('loggingEnabled', function (newVal, oldVal) {
+            if (newVal !== oldVal) {
+                $scope.isNotChanged = false;
+                if (newVal) {
+                    $scope.accessLogConfirmationDialog.foundation('reveal', 'open');
+                }
+            }
+        });
+        $scope.accessLogConfirmationDialog.on('opened.fndtn.reveal', function () {
+            $scope.accessLoggingConfirmed = false;
```
dkavanagh: Why not simply set this value above before showing the dialog?

dkavanagh:
hal, retest this please

eucaconsole/static/js/pages/elb_wizard.js
kamalgill: @dkavanagh: The value is set at the controller level as well.  The value here (in the dialog) is reset to false since a user may enable or disable access logging numerous times on the page (i.e. they can check/uncheck the "Enable logging" checkbox prior to clicking Save Changes), so we'd want the confirm dialog to display each time the "Enable logging" checkbox is checked.  At least that's how I interpreted the wireframe  -- we should confirm with @JMoLo if the confirmation dialog only needs to be displayed once per page load.

JMoLo: Now that you mention it, I think that dialog would be a great place for a "don't show me this again" checkbox....shall I add it to the wireframe?

kamalgill: @JMoLo : Yep, a "don't show me again" checkbox sounds like a good option.

JMoLo: Wireframe updated! https://eucalyptus.atlassian.net/wiki/display/3UD/Create+Load+Balancer

PR #1532, opened 2015-08-28
GUI-1954 - Update form field to text type, special filter on input

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-1954


euca-qa-gui-tester:

PR #1534, opened 2015-08-31
GUI-1957: Display reason for unhealthy instance status on ELB Instances page

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-1957

### DIFF
--- a/eucaconsole/static/html/directives/instance_selector.html
+++ b/eucaconsole/static/html/directives/instance_selector.html
@@ -1,4 +1,4 @@
-<div id="instance-selector" ng-cloak="" xmlns:i18n="http://www.w3.org/1999/xhtml">
+<div id="instance-selector" ng-cloak="">
     <table class="table">
         <thead>
             <tr>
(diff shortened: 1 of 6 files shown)


eucaconsole/static/html/directives/instance_selector.html
```diff
@@ -8,6 +8,7 @@
                 <th>{{ tableText.zone }}</th>
                 <th ng-if="isVPCSupported">{{ tableText.subnet }}</th>
                 <th>{{ tableText.status }}</th>
```
coyote240: Added class to column header to avoid reflow after the rest of the table data renders.

eucaconsole/static/html/directives/instance_selector.html
```diff
@@ -1,4 +1,4 @@
-<div id="instance-selector" ng-cloak="">
+<div id="instance-selector" ng-cloak="" xmlns:i18n="http://www.w3.org/1999/xhtml">
```
kamalgill: @coyote240: We can remove the i18n namespace attr here since this isn't a Chameleon XHTML template (i.e. it's a plain-old HTML file referenced by an Angular directive).

PR #1536, opened 2015-08-31
GUI-1966 Update layout of form fields and labels.

coyote240:
Set left and right large column widths on form fields and labels.
https://eucalyptus.atlassian.net/browse/GUI-1966

### DIFF
--- a/eucaconsole/templates/ipaddresses/ipaddresses.pt
+++ b/eucaconsole/templates/ipaddresses/ipaddresses.pt
@@ -98,7 +98,7 @@
             </metal:block>
         </div>
         <!--! Modal dialogs -->
-        <div id="allocate-ip-modal" class="reveal-modal small" data-reveal="">
+        <div id="allocate-ip-modal" class="reveal-modal medium" data-reveal="">
             <h3 i18n:translate="">Allocate elastic IP addresses</h3>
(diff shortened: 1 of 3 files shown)


euca-qa-gui-tester:

eucaconsole/templates/ipaddresses/ipaddresses.pt
```diff
@@ -121,7 +121,9 @@
                         <div tal:condition="is_vpc_supported" class="small-6 columns value" i18n:translate="">VPC</div>
                     </div>
                 </div>
-                ${panel('form_field', field=allocate_form.ipcount, leftcol_width=6, rightcol_width=6, **html_attrs)}
+                ${panel('form_field', field=allocate_form.ipcount,
```
kamalgill: @coyote240: There's a bug in the version of Chameleon in our target deployment environment that requires the ${panel...} calls to be on a single line, so we'll need to remove the line breaks here.  Once we move to CentOS 7 and Python 2.7, we can add the line breaks to the panel calls.

aliceh:
ok to test

aliceh:
test this please

coyote240:
Hal, test this please

aliceh:
Guitester, retest this please

coyote240:
Please test using both AWS and Euca as there are field type differences between the two.

PR #1537, opened 2015-08-31
GUI-1976: Better handling of missing error message when the ufshost is unreachable on login attempts

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-1976


dkavanagh:
looks good. thanks for improving that code

PR #1539, opened 2015-09-01
GUI-1967: Tighten CSP by setting form-action directive

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-1967


dkavanagh:
This seems a no-brainer to merge.

PR #1540, opened 2015-09-01
GUI-1969 Focus name field on page load.

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-1969

### DIFF
--- a/eucaconsole/templates/elbs/elb_wizard.pt
+++ b/eucaconsole/templates/elbs/elb_wizard.pt
@@ -49,7 +49,7 @@
                     <div class="tabs-content">
                         <!--! Step 1: General tab content -->
                         <div class="content" ng-class="stepClasses[0]" id="step1"
-                            tal:define="html_attrs_val {'class': 'text-input'};">
+                            tal:define="html_attrs_val {'class': 'text-input', 'focus-on-load': 'true'};">
                             <div ng-cloak="true">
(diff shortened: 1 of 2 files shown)


eucaconsole/templates/elbs/elb_wizard.pt
```diff
@@ -49,7 +49,7 @@
                     <div class="tabs-content">
                         <!--! Step 1: General tab content -->
                         <div class="content" ng-class="stepClasses[0]" id="step1"
-                            tal:define="html_attrs_val {'class': 'text-input'};">
+                            tal:define="html_attrs_val {'class': 'text-input', 'focus-on-load': 'true'};">
```
kamalgill: @coyote240: Your focusOnLoad directive is fairly slick, but we could use the HTML5 autofocus attribute here, right?  Meaning, the html_attrs_val could be `{'class': 'text-input', 'autofocus': 'autofocus'}` here, no Angular directive needed.

coyote240: Agreed, but Firefox in this case is touchy.  Because the "Step 1" div starts without the 'active' class and is cloaked, the div starts invisible and the autofocus is not honored.  Chrome works fine.

I guess we can either use the directive, or we can set active on the first "Step" and remove ng-cloak as it would appear that this is rendered entirely server-side.  I don't know the flow well enough to be sure, however.

Thoughts?

PR #1549, opened 2015-09-04
GUI-1986: Allow creation of an ELB without access logging enabled

kamalgill:
Bucket name field in Create ELB Wizard requires special validation handling since it is conditionally required.

Fixes https://eucalyptus.atlassian.net/browse/GUI-1986


eucaconsole/static/js/pages/elb_wizard.js
```diff
@@ -685,6 +685,18 @@ angular.module('BaseELBWizard').controller('ELBWizardCtrl', function ($scope, $h
         $scope.accessLoggingConfirmed = true;
         $scope.accessLogConfirmationDialog.foundation('reveal', 'close');
     };
+    $scope.createELB = function () {
```
dkavanagh: This call was added. What invokes it?

kamalgill: @dkavanagh: The submit button on the create ELB wizard had an `ng-click=createELB()`, which was added (by Kyo) a while ago but was never implemented.

PR #1554, opened 2015-09-08
GUI-1990: Opt out of PEP 476 based on config setting

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-1990

### DIFF
--- a/eucaconsole/views/__init__.py
+++ b/eucaconsole/views/__init__.py
@@ -147,6 +147,19 @@ class BaseView(object):
 
         return conn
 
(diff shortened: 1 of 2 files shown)


dkavanagh:
It seems like we could disable our cert checking in favor of the built-in checking, however we support supplying the cert file and we'd need to determine if we could override the standard one for this as well. Perhaps we could defer that till later.

dkavanagh:
Since this is a global ssl config variable, can't this just be done 1 place in config.py?

PR #1556, opened 2015-09-09
GUI-1971: Improved handling of empty chart messages

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-1971

### DIFF
--- a/eucaconsole/static/js/pages/cloudwatch_charts.js
+++ b/eucaconsole/static/js/pages/cloudwatch_charts.js
@@ -16,6 +16,7 @@ angular.module('CloudWatchCharts', ['EucaConsoleUtils'])
     vm.largeChartDuration = 3600;
     vm.largeChartGranularity = 300;
     vm.metricTitleMapping = {};
+    vm.emptyMessages = {};
     vm.chartsList = [];
     vm.granularityChoices = [];
     vm.originalDurationGranularitiesMapping = {};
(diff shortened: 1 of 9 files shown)


eucaconsole/static/js/pages/cloudwatch_charts.js
```diff
@@ -149,8 +161,10 @@ angular.module('CloudWatchCharts', ['EucaConsoleUtils'])
                 'HTTPCode_Backend_4XX', 'HTTPCode_Backend_5XX'
             ];
             var preciseFormatterMetrics = ['Latency'];
+            var leftMargin = 68;
+            var rightMargin = 38;
             var chart = nv.models.lineChart()
-                .margin({left: 68, right: 38})
+                .margin({left: leftMargin, right: rightMargin})
```
dkavanagh: Any reason to re-factor this since values are not re-used?

kamalgill: @dkavanagh: Good catch.  I had refactored this when tweaking the padding for the empty data message, but since those have been moved outside the SVG chart the refactoring is no longer necessary -- will simplify.

dkavanagh: I had a feeling. I've done similar things. Sometimes, examining the diff after the PR is created is helpful to me.

PR #1562, opened 2015-09-10
GUI-1994, GUI-1993 better shared bucket error handling, better double-slash handling.

dkavanagh:
catch errors in bucket access for shared buckets,
override boto's double-slash default behavior for bucket item details page

https://eucalyptus.atlassian.net/browse/GUI-1993
https://eucalyptus.atlassian.net/browse/GUI-1994


dkavanagh:
I have 3 unit tests to fix.

PR #1564, opened 2015-09-11
GUI-1565: Remove misleading warning when session.secure = false and console is requested via HTTPS

kamalgill:
...also set session.secure = true in the default console.ini file

Fixes https://eucalyptus.atlassian.net/browse/GUI-1565


dkavanagh:
hal, retest this please

kamalgill:
Hal, retest this please

aliceh:
From the log it is evident that the tester can not log in to the console. Not sure why...

dkavanagh:
code changes are good. I can also login when running it on OS X. I can't see how this would interfere with login, but would be nice to know why tests are failing.

PR #1572, opened 2015-09-14
GUI-1999 don't show drop-down if empty

dkavanagh:
don't show drop-down if empty

https://eucalyptus.atlassian.net/browse/GUI-1999

### DIFF
--- a/eucaconsole/static/js/thirdparty/magic-search/magic_search.html
+++ b/eucaconsole/static/js/thirdparty/magic-search/magic_search.html
@@ -20,7 +20,7 @@
             <div class="search-entry" is-open="isMenuOpen">
                 <input class="search-input" type="text" data-dropdown="facet-drop" dropdown-toggle=""
                        placeholder="{{ strings.prompt }}" autocomplete="off" />
-                <ul id="facet-drop" class="f-dropdown dropdown-menu" data-dropdown-content="">
+                <ul id="facet-drop" class="f-dropdown dropdown-menu" data-dropdown-content="" ng-show="filteredObj.length > 0">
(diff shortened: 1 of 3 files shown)


eucaconsole/static/js/thirdparty/magic-search/magic_search.html
```diff
@@ -20,7 +20,7 @@
             <div class="search-entry" is-open="isMenuOpen">
                 <input class="search-input" type="text" data-dropdown="facet-drop" dropdown-toggle=""
                        placeholder="{{ strings.prompt }}" autocomplete="off" />
-                <ul id="facet-drop" class="f-dropdown dropdown-menu" data-dropdown-content="">
+                <ul id="facet-drop" class="f-dropdown dropdown-menu" data-dropdown-content="" ng-show="filteredObj.length > 0">
```
kamalgill: @dkavanagh: We'll need to escape the `>' here to avoid breaking i18n

eucaconsole/templates/macros.pt
```diff
@@ -44,9 +44,9 @@
             </div>
         </div>
         <div class="lp-header">
-            <div class="lp-search">
+            <div class="lp-search" tal:define="prompt layout.searchtext_prompt if len(search_facets) > 2 else layout.searchtext_prompt2">
```
kamalgill: @dkavanagh: We'll need to escape the `>' here to avoid breaking i18n

eucaconsole/static/js/thirdparty/magic-search/magic_search.html
dkavanagh: not needed. Just the '<' is needed. (same w/ below)

PR #1584, opened 2015-09-18
GUI-2017 Mind the block device mapping, summary

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-2017


coyote240:
Hal, retest this please.

coyote240:
Hal, retest this please

aliceh:
retest this, please

eucaconsole/templates/instances/instance_launch.pt
```diff
@@ -310,6 +310,11 @@
                             <i class="fi-check"></i> Uses private addressing
                         </div>
                     </div>
+                    <div class="row">
+                        <div class="columns value" ng-show="additionalStorageConfigured">
+                            <i class="fi-check"></i> Additional storage configured
```
kamalgill: @coyote240: We'll need to mark the "Additional storage configured" string for i18n (ditto for line 310 above, while we're here).

coyote240: @kamalgill Done.

eucaconsole/static/js/widgets/bdmapping_editor.js
```diff
@@ -46,7 +50,16 @@ angular.module('BlockDeviceMappingEditor', ['EucaConsoleUtils'])
                 $scope.checkValidInput();
             });
 
-            $http.get("/instances/new/nextdevice/json").then(function (oData) {
+            $scope.$watch('bdMapping', function (newMapping) {
+                $scope.$emit('bdMappingChange', additionalStorageConfigured(newMapping));
+            });
+
+            var devicesMappings = Object.keys($scope.bdMapping);
```
dkavanagh: bdMapping isn't always defined, so this would throw an error in that case.

kamalgill: @dkavanagh: $scope.bdMapping is an empty object ({}) when unset, and Object.keys() simply returns an empty array in that case, so I think we're fine here.

dkavanagh: I only mentioned because I was getting an error in firebug's console.  "Error: can't convert undefined to object" at line 57 in the bdmapping_editor.js. I notice line 100 sets bdMapping to "undefined". Also line 108, so I think it can be undefined and perhaps these other places just need to be changed as well.

kamalgill: Yeah, good catch.  We should convert the `$scope.bdMapping = undefined` to `$scope.bdMapping = {}` in those two places.  Should we open a ticket for 4.2.1?  Does that break anything?
/cc @coyote240

coyote240: Could add a guard like this:
```
var deviceMappings = Object.keys($scope.bdMapping || {});
```

dkavanagh: It probably prevents the chosen from being initialized. Notice after "setInitialNewValues()" is called, "initChosenSelector()" would be called in the init code. I think we do need a ticket for 4.2.1.

PR #1585, opened 2015-09-18
GUI-2016 Utilize existing service to deliver suggested device mapping…

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-2016

Addresses bug by reusing an existing service.  Modified the service slightly to loosen the coupling just a bit and introduce a usage pattern where a new path may be recommended for an instance without an ID - a new instance.

### DIFF
--- a/eucaconsole/forms/instances.py
+++ b/eucaconsole/forms/instances.py
@@ -325,8 +325,11 @@ class AttachVolumeForm(BaseSecureForm):
         self.volume_id.error_msg = self.volume_error_msg
         self.device.error_msg = self.device_error_msg
         self.set_volume_choices()
+
         if self.instance is not None:
-            self.device.data = AttachVolumeForm.suggest_next_device_name(request, instance)
+            cloud_type = request.session.get('cloud_type')
+            mappings = instance.block_device_mapping
+            self.device.data = AttachVolumeForm.suggest_next_device_name(cloud_type, mappings)
 
     def set_volume_choices(self):
         """Populate volume field with volumes available to attach"""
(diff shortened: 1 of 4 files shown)


eucaconsole/forms/instances.py
```diff
@@ -341,20 +341,17 @@ def set_volume_choices(self):
         self.volume_id.choices = choices
 
     @staticmethod
-    def suggest_next_device_name(request, instance):
-        cloud_type = request.session.get('cloud_type')
+    def suggest_next_device_name(cloud_type, mappings):
```
coyote240: Loosening the coupling just a bit.  Generally best, IMHO, to pull the values off a request object than to push the request object deeper into the stack.

eucaconsole/forms/instances.py
```diff
@@ -341,20 +341,17 @@ def set_volume_choices(self):
...
             dev_root = '/dev/sd'
             start_char = 102
-        mappings = instance.block_device_mapping
+
         for i in range(0, 10):   # Test names with char 'f' to 'p'
             dev_name = dev_root+str(unichr(start_char+i))
-            try:
-                mappings[dev_name]
-            except KeyError:
+            if dev_name not in mappings:
```
coyote240: Just a bit more pythonic.

eucaconsole/forms/instances.py
kamalgill: @coyote240:  Since we've changed the method signature here, we'll also need to update line 329 in forms/instances.py...
 `self.device.data = AttachVolumeForm.suggest_next_device_name(request, instance)`

coyote240: Good catch, thanks!

PR #1587, opened 2015-09-18
GUI-2040: Fix typo in CSS border-color rule for split buttons on landing pages

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2040


eucaconsole/static/sass/includes/_landingpage.scss
```diff
@@ -475,7 +475,7 @@ $dropdown-border-color: $hp-gray-65;
 }
 
 .euca-split {
-    #border-color: $euca-extralightgrey !important;
```
dkavanagh: This should be removed entirely. Visually, there is no difference with the change in this PR

PR #1593, opened 2015-09-24
GUI-2039: For Create Key Pair on the Security tabs of Launch Instance…

lnajjar:
… and Launch Configuration, I changed the button label from "Create Key Pair" to "Create and Download."


euca-qa-gui-tester:

PR #1597, opened 2015-09-24
GUI-2086 add check before looking in dict

dkavanagh:
add check before looking in dict

https://eucalyptus.atlassian.net/browse/GUI-2086

### DIFF
--- a/eucaconsole/views/stacks.py
+++ b/eucaconsole/views/stacks.py
@@ -834,7 +834,7 @@ class StackWizardView(BaseView, StackMixin):
                 return  # ignore refs already in params
             if type(item) is dict and 'ImageId' in item.keys():
                 img_item = item['ImageId']
-                if 'Ref' not in img_item.keys():
+                if isinstance(img_item, dict) and 'Ref' not in img_item.keys():
                     # check for emi lookup in map
                     if 'Fn::FindInMap' in img_item.keys():
                         map_name = img_item['Fn::FindInMap'][0]


eucaconsole/views/stacks.py
```diff
@@ -834,7 +834,7 @@ def find_image_ref(_name, item):
                 return  # ignore refs already in params
             if type(item) is dict and 'ImageId' in item.keys():
                 img_item = item['ImageId']
-                if 'Ref' not in img_item.keys():
+                if isinstance(img_item, dict) and 'Ref' not in img_item.keys():
```
coyote240: Can we not ensure upstream that item['ImageId'] is a dict?  Also, recommend using item.get('ImageId').

dkavanagh: json.loads always creates dicts, AFAIK that and arrays.

kamalgill: @dkavanagh  json.loads doesn't always create dicts -- see https://docs.python.org/2/library/json.html
I'm with @coyote240 on using dict.get() when possible

dkavanagh: In response to the upstream checking, I don't think so. The JSON doc can some from many places and it is certainly valid to have a string here instead of a dict.

dkavanagh: I think the ImageId lookup here is just fine in context of the checks in the  line above it. We should re-evaluate this code more completely in one fell swoop and test more fully afterwards.

PR #1599, opened 2015-09-25
GUI-2082: Prevent unexpected 404 error when viewing object details page on Eucalyptus

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2082

### DIFF
--- a/eucaconsole/views/buckets.py
+++ b/eucaconsole/views/buckets.py
@@ -850,6 +850,7 @@ class BucketItemDetailsView(BaseView, BucketMixin):
         with boto_error_handler(request):
             if self.s3_conn and self.bucket is None:
                 self.bucket = BucketContentsView.get_bucket(request, self.s3_conn)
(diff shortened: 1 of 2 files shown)


eucaconsole/views/buckets.py
```diff
@@ -850,7 +850,6 @@ def __init__(self, request, bucket=None, bucket_item_acl=None, **kwargs):
         with boto_error_handler(request):
             if self.s3_conn and self.bucket is None:
                 self.bucket = BucketContentsView.get_bucket(request, self.s3_conn)
-                self.s3_conn.suppress_consec_slashes = False
```
dkavanagh: This is required! Otherwise boto will squash paths that have "//" in them. Let's look at why this fails, not just remove it.

kamalgill:
Closing w/o merging since the root cause was a ufshost misconfiguration

PR #1601, opened 2015-09-25
GUI-2093 don't try to modify tags prefixed with aws:, euca: or unmodified

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2093

### DIFF
--- a/eucaconsole/views/scalinggroups.py
+++ b/eucaconsole/views/scalinggroups.py
@@ -339,14 +339,15 @@ class ScalingGroupView(BaseScalingGroupView, DeleteScalingGroupMixin):
         return self.render_dict
 
     def update_tags(self):
-        if self.scaling_group.tags:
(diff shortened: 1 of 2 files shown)


eucaconsole/views/scalinggroups.py
```diff
@@ -338,12 +339,43 @@ def scalinggroup_delete(self):
         return self.render_dict
 
     def update_tags(self):
+        if self.scaling_group.tags:
+            # cull tags that start with aws: or euca:
+            self.scaling_group.tags = [
```
kamalgill: @dkavanagh: We should change this to a local var (e.g. scaling_group_tags) instead of modifying the scaling group tags in-place.

dkavanagh: I thought about that, but since this is the last thing the view does before a different view is shown, the object gets thrown away anyway.

PR #1607, opened 2015-09-29
GUI-2088 Added warnings if ufshost configured for localhost/127.0.0.1

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2088

### DIFF
--- a/eucaconsole/templates/dialogs/ufshost_warn_dialog.pt
+++ b/eucaconsole/templates/dialogs/ufshost_warn_dialog.pt
@@ -1,7 +1,7 @@
 <!--! Modal dialog to warn about ufshost configuration error -->
 <div tal:omit-tag="" i18n:domain="eucaconsole">
     <div id="ufshost-warn-dialog" class="reveal-modal small" data-reveal="">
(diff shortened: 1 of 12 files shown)


eucaconsole/templates/dialogs/ufshost_warn_dialog.pt
```diff
@@ -0,0 +1,16 @@
...
+    <div id="ufshost-warn-dialog" class="reveal-modal small" data-reveal="">
+        <h3 class="warn-dialog" i18n:translate="">WARNING: Cannot create stack</h3>
+        <p i18n:translate="">
+            You cannot create a stack because the ufshost setting includes localhost.
+        </p>
+        <p i18n:translate="">
+            Please ask your administrator to change the ufshost setting to the fully qualified domain name or the IP address of the machine running User Facing Services.
+        </p>
+        <div>
+            <a ng-click="closeDialog('ufshost-warn-dialog')" id="close_button" class="button expand" i18n:translate="">Close</a>
```
kamalgill: @dkavanagh IDs must be unique per page, so we should be more specific here than `id="close_button"`

PR #1621, opened 2015-10-06
GUI-2141: Better regex for parsing the ELB health check ping path from the target

kamalgill:
Fixes:
https://eucalyptus.atlassian.net/browse/GUI-2141


eucaconsole/forms/elbs.py
```diff
@@ -250,11 +250,11 @@ def set_health_check_choices(self):
 
     def get_health_check_data(self):
         if self.elb is not None and self.elb.health_check.target is not None:
-            match = re.search('^(\w+):(\d+)/?(.+)?', self.elb.health_check.target)
+            match = re.search('^(\w+):(\d+)(/)?(.+)?', self.elb.health_check.target)
```
dkavanagh: Should the "/" be escaped? I tried on regex101.com and it indicated I should do that "^(\w+):(\d+)(\/)?(.+)?"

dkavanagh: I wonder, shouldn't we always expect the slash? ^(\w+):(\d+)\/(.+)? I guess it depends on where we're checking. If we're checking the API Target value, always one slash after the port, but if we're checking the input field, that's different. (though this regex contains more than just the path)

kamalgill: @dkavanagh: We don't need to escape "/" in a python regex.

Yeah, good point about expecting the slash for HTTP/HTTPS targets.  I wonder if we should change the regex to `^(\w+):(\d+)(.+)`.  If we leave it as `^(\w+):(\d+)(/)?(.+)?`, the ping path for an ELB with a health check target of `HTTP:80/weather/us/wa/seattle` would be parsed with a ping path of `weather/us/wa/seattle`, but I think we'd want it to be parsed as `/weather/us/wa/seattle` (which should work with the new regex).

dkavanagh:
could you also set the client-side validation to ensure the leading slash? I'd be happy then!

kamalgill:
@dkavanagh: Updating the client-side validation is a bit risky this late in the release cycle, so I opted to add a leading slash to the ping path server side when necessary.

kamalgill:
Hal, retest this please

kamalgill:
Hal, retest this please

kamalgill:
Hal, retest this please

kamalgill:
Hal, retest this please

aliceh:
I tested packages created from this branch against setup#1 cloud as backend. All automated tests passed. @kamalgill this PR is good to merge.

PR #1624, opened 2015-10-08
GUI-2103 updated text to determine if 403 is due to permissions

dkavanagh:
updated text to determine if 403 is due to permissions

https://eucalyptus.atlassian.net/browse/GUI-2103


kamalgill:
Hal, retest this please

PR #1635, opened 2015-10-12
DOC-1687: updated launch config from instance and new file launch con…

tvo318:
…fig more like this. 
Please see GUI-2151.


kamalgill:
https://eucalyptus.atlassian.net/browse/GUI-2157

PR #1642, opened 2015-10-13
Pin M2Crypto to avoid build issue with 0.22.4

kamalgill:
Will need to revert when upstream fixes, per https://eucalyptus.atlassian.net/browse/GUI-2161


gholms:
Upstream is already aware of this, so if you merge this change please revert it once they fix the bug.  https://gitlab.com/m2crypto/m2crypto/issues/60

PR #1650, opened 2015-10-20
GUI-1554

dkavanagh:
cloud formation template we can link to from docs

### DIFF
--- a/scripts/console-deploy.cfn
+++ b/scripts/console-deploy.cfn
@@ -1,0 +1,143 @@
+{
+  "AWSTemplateFormatVersion" : "2010-09-09",
+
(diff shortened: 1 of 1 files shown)


scripts/console-deploy.cfn
```diff
@@ -0,0 +1,143 @@
...
+      "Properties" : {
+        "GroupDescription" : "Enable SSH access and HTTP from the load balancer only",
+        "SecurityGroupIngress" : [ {
+          "IpProtocol" : "tcp",
+          "FromPort" : "22",
+          "ToPort" : "22",
+          "CidrIp" : { "Ref" : "SSHLocation"}
+        },
+        {
+          "IpProtocol" : "tcp",
```
kamalgill: @dkavanagh: shouldn't the protocol be 'http' here?

PR #1655, opened 2015-10-22
GUI-2168 Prevent user from trying to create an admin user, update val…

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-2168


kamalgill:
@coyote240: GUI-2168 is scoped for 4.2.1, so we'll need to branch off (and open a PR against) maint-4.2 rather than master

PR #1663, opened 2015-10-23
Gui 1993 modify xhr view connection to stop suppressing consec slashes

dkavanagh:
modify xhr view connection to stop suppressing consec slashes

https://eucalyptus.atlassian.net/browse/GUI-1993


dkavanagh:
hal, retest this please

PR #1665, opened 2015-10-23
GUI-2175 handle different protocols for each side of listener

dkavanagh:
handle different protocols for each side of listener

https://eucalyptus.atlassian.net/browse/GUI-2175


kamalgill:
Hal, retest this please

PR #1672, opened 2015-10-24
GUI-2177: Bump down bucket icon size in tile view on buckets landing page

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2177


kamalgill:
Hal, retest this please

PR #1679, opened 2015-10-27
GUI-2159: Disable "Create Load Balancer" button in ELB wizard when access logging is enabled and a bucket isn't selected or created

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2159


kamalgill:
Hal, retest this please

PR #1681, opened 2015-10-27
Gui 2154 populate user data for "launch config like" and "launch more"

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2154

### DIFF
--- a/eucaconsole/forms/instances.py
+++ b/eucaconsole/forms/instances.py
@@ -28,11 +28,12 @@
 Forms for Instances
 
 """
+import base64
 import wtforms
 from wtforms import validators
 
 from ..i18n import _
-from ..views import BaseView
+from ..views import BaseView, boto_error_handler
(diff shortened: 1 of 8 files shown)


eucaconsole/forms/instances.py
```diff
@@ -269,6 +270,10 @@ def set_initial_data(self):
         self.monitoring_enabled.data = self.instance.monitored
         self.private_addressing.data = self.enable_private_addressing()
         self.number.data = 1
+        with boto_error_handler(self.request):
+            userdata = self.conn.get_instance_attribute(self.instance.id, 'userData')
+            userdata = userdata['userData']
```
kamalgill: @dkavanagh: Does the userdata dict always have the 'userData' key, or should we change this to be `userdata = userdata.get('userData')` to be safe?

dkavanagh: In testing, I found that even when there wasn't any user data, we still had the userData wrapper.

PR #1682, opened 2015-10-27
Gui 1925 don't filter snapshot request when volumes exceed 1000

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1925


kamalgill:
Hal, retest this please

PR #1685, opened 2015-10-28
GUI-2041: Set ELB health check protocol/port from first instance protocol/port entry in listeners list

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2041

### DIFF
--- a/eucaconsole/static/js/pages/elb_wizard.js
+++ b/eucaconsole/static/js/pages/elb_wizard.js
@@ -233,9 +233,15 @@ angular.module('BaseELBWizard').controller('ELBWizardCtrl', function ($scope, $h
                 $scope.updateAvailabilityZoneChoices();
             }
         }, true);
-        $scope.$watch('pingProtocol', function (){
-            $scope.updateDefaultPingProtocol();
-        });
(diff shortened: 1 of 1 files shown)


kamalgill:
Hal, retest this please

eucaconsole/static/js/pages/elb_wizard.js
```diff
@@ -233,9 +233,15 @@ angular.module('BaseELBWizard').controller('ELBWizardCtrl', function ($scope, $h
                 $scope.updateAvailabilityZoneChoices();
             }
         }, true);
-        $scope.$watch('pingProtocol', function (){
-            $scope.updateDefaultPingProtocol();
-        });
+        $scope.$watch('listenerArray', function (newVal, oldVal) {
```
dkavanagh: Functionally, does this listener always adjust the HC protocol and port if a listener has been added? What if a user has set protocol/port on last step then gone back to add another listener?

kamalgill: The protocol/port will be updated if the user goes back and adds another listener.

PR #1698, opened 2015-11-06
GUI-2174: Properly configure bucket access ACLs even when the prefix is blank

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2174


dkavanagh:
Didn't we need to provide access to the eucalyptus loadbalancer user? (per Sang-Min)

kamalgill:
@dkavanagh: We're already setting the proper bucket ACLs in the configure_logging_bucket_acl method, but that wasn't called when the prefix was empty.

kamalgill:
Jenkins build has passed ... http://jenkins.qa1.eucalyptus-systems.com/view/Eucaconsole%20CI/job/user-console-PR-monitor/1136/

PR #1801, opened 2016-02-17
GUI-2257: Prevent validation error for unit field when creating alarms for ASG metrics

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2257 (for 4.2.2)

### DIFF
--- a/eucaconsole/constants/cloudwatch.py
+++ b/eucaconsole/constants/cloudwatch.py
@@ -32,6 +32,7 @@ from ..i18n import _
 
 
 METRIC_TYPES = [
+    # NOTE: AWS/AutoScaling metrics expect 'None' (not 'Count' or None) as the unit
     {'namespace': 'AWS/AutoScaling', 'name': 'GroupDesiredCapacity', 'unit': 'None'},
     {'namespace': 'AWS/AutoScaling', 'name': 'GroupInServiceInstances', 'unit': 'None'},
     {'namespace': 'AWS/AutoScaling', 'name': 'GroupMaxSize', 'unit': 'None'},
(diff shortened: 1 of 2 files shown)


euca-qa-gui-tester:
Refer to this link for build results (access rights to CI server needed): 
http://jenkins.qa1.eucalyptus-systems.com//job/user-console-PR-monitor/1516/
Test FAILed.

eucaconsole/constants/cloudwatch.py
```diff
@@ -32,6 +32,7 @@
 
 
 METRIC_TYPES = [
+    # NOTE: AWS/AutoScaling metrics expect 'None' (not 'Count' or None) as the unit
```
coyote240: Just because there has been contention, can you provide a reference link for this?  Also, is the expectation the same for alarms?

eucaconsole/forms/alarms.py
```diff
@@ -187,7 +187,7 @@ def get_statistic_choices():
 
     @staticmethod
     def get_unit_choices():
-        choices = [BLANK_CHOICE]
+        choices = [BLANK_CHOICE, ('None', 'None')]
```
dkavanagh: Is "None" a valid choice for non ASG metrics?

kamalgill: @dkavanagh: Good question -- I don't know if any non-ASG metrics allow 'None' as a unit, and I can't find any AWS docs on the topic. http://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDatum.html doesn't break down metrics by unit, and the boto docs don't either.

At any rate, the list here is for all possible unit choices, regardless of metric/namespace.  The appropriate unit is selected based on the metric-unit mapping in the METRIC_TYPES constant. Revisiting the create alarm code again, we probably should have automatically mapped the unit in the create_alarm view rather than building a form field (with restricted choices) for it.  It's probably too risky to put that change in a bugfix release, but we could refactor that later.

PR #1822, opened 2016-03-02
GUI-2360 implement stack update

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2360

PR #1832, opened 2016-03-08
GUI-2360: Implement stack update

kamalgill:
https://eucalyptus.atlassian.net/browse/GUI-2360
Supplants https://github.com/eucalyptus/eucaconsole/pull/1822 due to errant commit there


aliceh:
Retest this please

eucaconsole/views/stacks.py
```diff
@@ -85,6 +95,29 @@ def get_create_template_bucket(self, create=False):
             u'Please contact your cloud administrator.'
         ))
 
+    def get_template_location(self, stack_id, default_name=None):
+        bucket = None
+        try:
+            bucket = self.get_create_template_bucket()
```
kamalgill: @dkavanagh: get_create_template_bucket() modifies state, which is a serious concern as the 'stack_view' route handler calls get_template_location().  Also, although it isn't part of this PR, the 'stack_view' config is missing a request_method arg (should be 'GET', right?), which defaults to all methods when unspecified according to the Pyramid docs at http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/viewconfig.html

eucaconsole/forms/stacks.py
```diff
@@ -57,6 +57,18 @@ def __init__(self, request, s3_bucket, **kwargs):
...
 
+class StacksUpdateForm(BaseSecureForm):
+    """Stacks update form.
+       Only need to initialize as a secure form to generate CSRF token
+    """
+    template_file_helptext = _(u'Template file may not exceed 16 KB')
+    template_file = wtforms.FileField(label='')
+
+    def __init__(self, request, s3_bucket, **kwargs):
+        super(StacksCreateForm, self).__init__(request, **kwargs)
```
kamalgill: @dkavanagh: This line should be `super(StacksUpdateForm, self).__init__(request, **kwargs)`

eucaconsole/static/js/pages/stacks.js
```diff
@@ -5,12 +5,13 @@
  */
 
 // Pull in common landing page module
-angular.module('StacksPage', ['LandingPage', 'EucaConsoleUtils'])
+angular.module('StacksPage', ['LandingPage', 'EucaConsoleUtils', 'angular.filter'])
```
kamalgill: @dkavanagh: Where are we using angular.filter?

eucaconsole/static/sass/pages/stack_update.scss
```diff
@@ -0,0 +1,73 @@
...
+    margin-left: $template-indent;
+    width: 90%;
+}
+
+small.error {
+    margin-left: $template-indent;
+}
+
+.error label {
+    color: $euca-grey;
```
kamalgill: @dkavanagh: This should probably be one of the $hp-gray values, as $euca-grey will probably go away in the rebranding effort.

eucaconsole/views/directives.py
```diff
@@ -1,6 +1,7 @@
 from pyramid.view import view_config
```
kamalgill: @dkavanagh: This file is missing the license header

eucaconsole/views/stacks.py
```diff
@@ -153,7 +187,18 @@ def __init__(self, request):
     def stacks_json(self):
         if not(self.is_csrf_valid()):
             return JSONResponse(status=400, message="missing CSRF token")
-        transitional_states = ['CREATE_IN_PROGRESS', 'ROLLBACK_IN_PROGRESS', 'DELETE_IN_PROGRESS', 'CREATE_FAILED']
+        transitional_states = [
```
kamalgill: @dkavanagh: Should we move this list to a constant -- somewhere more reusable?  Is it required anywhere else?

eucaconsole/views/stacks.py
```diff
@@ -423,6 +444,7 @@ def get_controller_options_json(self):
         return BaseView.escape_json(json.dumps({
             'stack_template_url': self.request.route_path('stack_template_parse'),
             'convert_template_url': self.request.route_path('stack_template_convert'),
+            'stack_template_read_url': self.request.route_path('stack_template', name=self.stack.stack_name) if self.stack else '',
```
kamalgill: @dkavanagh: We should wrap this line for our line length compliance

eucaconsole/views/stacks.py
dkavanagh: I think the concern over "get_create_template_bucket() is unfounded as that operation is idempotent. If, at any point, the bucket was created, it will never be again. We must have that location in the view. I went through the thinking on this months ago when struggling about the inclusion of this in the stack_view.

eucaconsole/static/js/pages/stacks.js
dkavanagh: Yes, and here's one of 2 instances: https://github.com/eucalyptus/eucaconsole/blob/GUI-2360/eucaconsole/templates/stacks/stacks.pt#L30

dkavanagh:
f'k.. I've been pushing code to GUI-2360 for the simple reason, that's the branch this is supposed to be on! No guarantees I'll have time to do the fixes on the GUI-2360-good branch.. or, perhaps I can reset the HEAD on GUI-2360 to ditch the errant commit. bah

eucaconsole/views/stacks.py
kamalgill: @dkavanagh: Sorry, we simply can't allow a GET request to create a bucket -- it's as simple as that.  If we need the bucket created, we should somehow alert the user and create the bucket via a user-generated POST.

dkavanagh: I believe we did add text in the UI for this.

https://github.com/eucalyptus/eucaconsole/blob/master/eucaconsole/templates/stacks/stack_wizard.pt#L65
https://github.com/eucalyptus/eucaconsole/blob/GUI-2360/eucaconsole/templates/stacks/stack_update.pt#L50

dkavanagh: It's actually the definition of idempotent if it were creating the bucket, but looking at the code more, it's not creating it on the view method. get_create_template_bucket is being called without setting create=True, so it's not going to do the create from the GET method.
@kamalgill didn't you add some code to the stack wizard to prompt if we created the bucket? Is that's what is really missing here?

kamalgill: @dkavanagh: Understood -- perhaps we added that in 4.2, I'll take a look to confirm

PR #1833, opened 2016-03-09
support for el6 and el7

mbacchi:


setup.py
```diff
@@ -134,7 +134,7 @@ def make_release_tree(self, base_dir, files):
     'pyramid >= 1.4',
     'pyramid_beaker >= 0.8',
     'pyramid_chameleon >= 0.1',
-    'pyramid_layout <= 0.8',
+    'pyramid_layout >= 0.8',
```
dkavanagh: @kamalgill do you remember if there's a reason we capped this at 0.8?

kamalgill: @dkavanagh: https://github.com/eucalyptus/eucaconsole/commit/d68a8e456fcfc9fcac4f5f58d12443dc9d520746

PR #1834, opened 2016-03-09
GUI-2464: Display alarm status on instances landing page

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2464

### DIFF
--- a/eucaconsole/templates/instances/instances.pt
+++ b/eucaconsole/templates/instances/instances.pt
@@ -185,7 +185,9 @@
                 </div>
                 <div ng-show="item.alarm_status">
                     <span class="label" title="CloudWatch alarm status" i18n:attributes="title" data-tooltip="">CW</span>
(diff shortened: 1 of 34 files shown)


eucaconsole/templates/instances/instances.pt
```diff
@@ -252,9 +263,21 @@
                     <span class="label radius status {{ item.status }}">{{ item.status }}</span>
                     <span class="busy" ng-show="item.transitional"></span>
                 </td>
+                <td class="alarm-status">
+                    <a class="label radius alarm-status {{ item.alarm_status.replace(' ', '_').toLowerCase() }}"
```
coyote240: Rather than write into the class attribute, use ng-class.  In this case, since it's non-conditional, simply put the expression into the ng-class.  Same with the above, consider using a filter.

PR #1835, opened 2016-03-09
GUI 2467 Implement alarm view w/ delete for instance detail page

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2467


kamalgill:
@dkavanagh: The templates folder is not a python package, so I'm against that approach.  Please revert your commit and use what's in the Pyramid docs I referenced.

PR #1841, opened 2016-03-14
GUI-2463: Display alarm status on load balancers landing page

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2463


kamalgill:
Retest this please

eucaconsole/static/js/pages/custom_filters.js
```diff
@@ -50,4 +50,10 @@ angular.module('CustomFilters', [])
         input = input.replace(sub_delims, '');
         return input.toLowerCase();
     };
-});
+})
+.filter('toClassName', function () {
+    return function (input) {
+        return input.replace(' ', '_').toLowerCase();
```
dkavanagh: fwiw, I've seen a number of cases where there is status containing multiple words and a single replace may work for alarm status, but not in the general sense as this filter is intended for. Using a global replace would be better for a generic filter like this:
 input.replace(/ /g, '_').toLowerCase();

Note, my use case replaced a "-" with a space, so the space above may need to be escaped.

kamalgill: @dkavanagh : Good catch -- will fix

eucaconsole/static/js/pages/custom_filters.js
```diff
@@ -53,7 +53,7 @@ angular.module('CustomFilters', [])
 })
 .filter('toClassName', function () {
     return function (input) {
-        return input.replace(' ', '_').toLowerCase();
+        return input.replace(/\s+/g, '_').toLowerCase();
```
dkavanagh: Do you really want to match 1 or more whitespace characters? using /\ /g would be consistent with the intent here.

kamalgill: @dkavanagh: Yeah, I see your point -- we'd probably never have tab chars in the input, for example.  Will simplify.

PR #1864, opened 2016-03-28
Gui 1922 rework and re-branding of navigation

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-1922


dkavanagh:
This includes commits from the GUI-2205 branch so that consistent branding would be used. That branch probably should be merged first.

eucaconsole/templates/panels/top_nav.pt
```diff
@@ -1,129 +1,122 @@
 <!-- top nav links (reused in off-canvas menu) -->
 <ul class="resources-nav left-align ${'off-canvas-list' if off_canvas else ''}" i18n:domain="eucaconsole">
-    <li> 
-        <a id="resource-menuitem-dashboard" href="${request.route_path('dashboard')}">
+    <li class="dashboard">
```
kamalgill: @dkavanagh: We're duplicating the CSS class at a parent and child level here and elsewhere.  Is that intentional?

dkavanagh: That's a marker class more than an actual style. It let's the javascript find it so the "active" class can be added where needed.

kamalgill: @dkavanagh: Ah, ok.  Perhaps we should use data-* attributes there instead, since the duplicated CSS classes are confusing and may cause display issues if/when we style those later?

kamalgill:
Jenkins is failing in snapshots sequence on create snapshot test (unable to find volume chosen field), which has nothing to do with the changes here.  Merging...

PR #1869, opened 2016-03-29
GUI-2498 additional check for undefined default region

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2498

### DIFF
--- a/eucaconsole/layout.py
+++ b/eucaconsole/layout.py
@@ -96,7 +96,7 @@ class MasterLayout(object):
                     self.has_regions = False
                 except socket.error:
                     self.has_regions = False
-        if self.has_regions:
+        if hasattr(self, 'regions'):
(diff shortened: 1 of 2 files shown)


ccassler:
Retest this please

aliceh:
The Snapshot sequence is failing, because the locator "css = '[class="chosen-single"]>span" for selecting volumes in "Create Snapshot" dialog was changed.  Please change the locator to "css=#volume_id_chosen>a" for the snapshot test to pass.

ccassler:
@aliceh: This test was having other issues.  The GUI wasn't coming up at all.  Instead, we were seeing Internal Server Error which I noted in the ticket.

aliceh:
@ccassler  if the console would not have been up, then the test would fail at login.

kamalgill:
Retest this please

PR #1874, opened 2016-04-01
POC, build create alarm dialog and dependencies into one file.

coyote240:
Experiment with script combining.  No minification, but that should be added.


dkavanagh:
If there a way to automate this, so we didn't have to manually do it for all pages.

coyote240:
That tends to be where RequireJS comes in.  In this case, I simply concatenated the components necessary for CreateAlarmModule, so it's targeted to a component.
One thing we will need to do is add this task to the watch, like we do with sass.

PR #1878, opened 2016-04-04
Gui 2466 create alarm from metrics landing.

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2466

### DIFF
--- a/eucaconsole/static/js/pages/metrics.js
+++ b/eucaconsole/static/js/pages/metrics.js
@@ -251,7 +251,22 @@ angular.module('MetricsPage', ['LandingPage', 'CloudWatchCharts', 'EucaConsoleUt
             }, 500);
         };
         vm.showCreateAlarm = function(metric) {
-            $scope.metricForAlarm = metric;
+            if (Array.isArray(chart)) {
+                var dims = [];
+                chart.forEach(function(val) {
+                    var tmp = {};
+                    val.resources.forEach(function(res) {
+                        tmp[res.res_type] = res.res_id;
+                    });
+                    dims.push(tmp);
+                });
+                var met = Object.copy(metric);
+                met.dimensions = dims;
+                $scope.metricForAlarm = met;
+            }
+            else {
+                $scope.metricForAlarm = metric;
+            }
             ModalService.openModal('createAlarm');
         };
     })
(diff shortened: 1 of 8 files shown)


dkavanagh:
hal, retest this please

dkavanagh:
hal, retest this please

dkavanagh:
hal, retest this please

aliceh:
@dkavanagh, there is a merge conflict in this branch. CI will not be able to run until the conflict is resolved.

dkavanagh:
Pretty fatal errors across the board in CI now!

tvo318:
Yikes!
-tvo

aliceh:
@tvo318 @dkavanagh  if you look in the log, the error is clearly caused by network connectivity problems. Kyle was doing maintenance on Saturday and VPN was down, could be related to that, or there could be something going on with git.

aliceh:
Retest this please

aliceh:
Retest this please

aliceh:
Retest this please

dkavanagh:
current failure is in scaling groups test. Failure to click (though it was found) '#nextstep-scalinggroup-modal .close-reveal-modal'
I couldn't reproduce l locally due to other failure to find element in left-nav (element is present and this test failed fast, so timeout may be super short).

aliceh:
@dkavanagh I was also looking at this failure. The click method has a two minute timeout, so I don't think the problem was with the test code. I have seen this in stack trace: 
"  raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.WebDriverException: Message: Session ID is null. Using WebDriver after calling quit()?"
This seems  to me an indication of a webdriver problem. Somehow the session must have quit during test.

aliceh:
I think this is good to merge.

eucaconsole/static/js/pages/metrics.js
```diff
@@ -250,6 +250,36 @@ angular.module('MetricsPage', ['LandingPage', 'CloudWatchCharts', 'EucaConsoleUt
                 $(".metrics-url-field").select();
             }, 500);
         };
+        vm.showCreateAlarm = function(metric) {
+            var dims = {}; 
+            if (Array.isArray(metric)) {  // called from chart actions
```
coyote240: This is effectively duplicate code.  Might be easier to maintain to transform the input first, then perform the function.
Otherwise, iff showCreateAlarm is the public interface, nothing wrong with requiring the input to have a specific shape.

dkavanagh: I'm now thinking of something where a non-array metric argument is simply wrapped so that they're all handled as arrays.

PR #1879, opened 2016-04-05
GUI-2505 - Default metrics in Create Alarm Dialog/Metrics Service

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-2505


eucaconsole/static/js/services/metric-service.js
```diff
@@ -13,8 +13,13 @@ angular.module('MetricServiceModule', ['EucaRoutes'])
                         method: 'GET',
                         url: path
                     }).then(function (result) {
+                        var metrics;
                         if(result && result.data) {
-                            _metrics[value] = result.data.metrics;
+                            metrics = result.data.metrics || [];
```
coyote240: Since there's a slash (/) in the namespace value, it's kinda a bummer to pass it in the URL.  Here, I'm going ahead and filtering client-side to avoid the hassle.

eucaconsole/templates/cloudwatch/create_alarm_modal_template.pt
```diff
@@ -34,8 +34,12 @@
...
         </div>
         <div class="small-3 columns">
-            <select name="statistic" ng-model="alarm.statistic"
-                                     ng-options="statistic for statistic in alarm.metric.statistics track by statistic">
+            <select name="statistic" ng-model="alarm.statistic">
+                <option>Minimum</option>
+                <option>Maximum</option>
+                <option>Sum</option>
+                <option>Average</option>
+                <option>SampleCount</option>
```
kamalgill: This should display as "Sample Count", so we'll need to separate the label from the value for this choice.  And if we want i18n, we should include the values for the other choices as well.

eucaconsole/views/metrics.py
```diff
@@ -366,6 +366,7 @@ def metrics_available_for_resource(self):
             'unit': next(metric['unit'] for metric in METRIC_TYPES if metric['name'] == m.name),
             'dimensions': m.dimensions,
             'name': m.name,
+            'label': _(METRIC_LABEL.get(m.name, m.name)),
```
kamalgill: @coyote240: We should use constants.cloudwatch.METRIC_TITLE_MAPPING since that has i18n-enabled labels.

coyote240: They're different data sets, which is correct?  METRIC_LABEL is in the file where it is used.

kamalgill: @coyote240 Ah, right, we have GUI-2493 to reconcile those distinctions, so we can worry about that later.

aliceh:
Console throws Internal server error, that's why the tests fail.

coyote240:
@aliceh I'm having a hard time finding the error.  Can you share?

eucaconsole/views/metrics.py
```diff
@@ -363,11 +331,20 @@ def metrics_available_for_resource(self):
...
 
         result = [{
-            'statistics': m.Statistics,
+            'index': [metric['name'] for metric in METRIC_TYPES].index(m.name),
             'unit': next(metric['unit'] for metric in METRIC_TYPES if metric['name'] == m.name),
             'dimensions': m.dimensions,
             'name': m.name,
+            'label': METRIC_TITLE_MAPPING.get(m.name, m.name),
             'namespace': m.namespace} for m in metrics]
+        result.sort(lambda a, b: cmp(a.get('index', 0), b.get('index', 0)))
```
dkavanagh: like me some lambda functions

eucaconsole/templates/cloudwatch/create_alarm_modal_template.pt
```diff
@@ -34,13 +34,17 @@
...
         </div>
         <div class="small-3 columns">
-            <select name="statistic" ng-model="alarm.statistic"
-                                     ng-options="statistic for statistic in alarm.metric.statistics track by statistic">
+            <select name="statistic" ng-model="alarm.statistic">
+                <option value="Average" i18n:translate="">Average</option>
+                <option value="Minimum" i18n:translate="">Minimum</option>
+                <option value="Maximum" i18n:translate="">Maximum</option>
+                <option value="Sum" i18n:translate="">Sum</option>
+                <option value="SampleCount" i18n:translate="">Data samples</option>
```
dkavanagh: Any reason it isn't built like this from server-side def? https://github.com/eucalyptus/eucaconsole/blob/master/eucaconsole/templates/dialogs/cloudwatch_chart_dialog.pt#L14-L16

coyote240: This came out of a discussion with @kamalgill and @JMoLo.  The label names are changing, and the values are static.  Hard-coding in HTML seemed appropriate.

PR #1883, opened 2016-04-06
GUI-1659 updated beaker to at least 1.7.0 and added pycryptopp depend…

dkavanagh:
…ency

https://eucalyptus.atlassian.net/browse/GUI-1659

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -118,7 +118,7 @@ class sdist_with_git_version(sdist):
         os.rename(new_ver_name, old_ver_name)
 
 requires = [
-    'beaker >= 1.5.4',
+    'beaker >= 1.7.0',
     'boto >= 2.38.0',
     'chameleon >= 2.5.3',
     'defusedxml >= 0.4',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -130,6 +130,7 @@ def make_release_tree(self, base_dir, files):
     'M2Crypto >= 0.20.2',
     'markupsafe >= 0.9.2',
     'pycrypto >= 2.0.1',
+    'pycryptopp >= 0.7.1',
```
kamalgill: @dkavanagh: I don't think we need pycrypto if we're using pycryptopp, per https://github.com/bbangert/beaker/blob/master/beaker/crypto/pycrypto.py

dkavanagh: is Crypto being imported, pycrypto?

kamalgill: Yes, AFAICT.  It's probably harmless to leave pycrypto there since Beaker falls back to pycrypto only when pycryptopp isn't there.

kamalgill:
Supplanted by #1928

PR #1884, opened 2016-04-07
GUI-2504: Layout fixes for Create Alarm dialog

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2504


aliceh:
Good to merge. EIP test failed because of my log archiving experimentation.

eucaconsole/static/js/widgets/modal.js
```diff
@@ -34,8 +34,10 @@ angular.module('ModalModule', [])
         if(!modal) {
             return;
         }
-
         modal.addClass('open');
+        modal.find('.modal-bg').on('click', function() {
```
coyote240: This will add a new click event every time the modal is opened.  Also, find() has only limited support in jqLite, so is worth avoiding as we may be able to remove the need for jQuery in the future.
However, I have tested adding:
`ng-click="closeModal(\'{{modalName}}\')`
to the modal-bg div and removing this find entirely with success.

eucaconsole/static/sass/widgets/modal.scss
```diff
@@ -1,38 +1,43 @@
+$modal-open-top-position: -40px;
+
 
 @mixin modal-width($width:60%) {
     .modal-content {
         width: $width;
     }
 }
 
+@keyframes slidedown {
```
coyote240: This is cool.

coyote240:
The amount field of the dialog is marked invalid when the form opens.  I would recommend adding `.ng-touched` and possibly `.ng-submitted` to `form .ng-invalid` in eucaconsole.scss (line 845?)

PR #1885, opened 2016-04-07
GUI-2516 converted volumes call to use filters

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2516


aliceh:
Volume test ran successfully, but shows "Failed" because I was modifying templates during that test and it could not collect one of the logs.  @dkavanagh, This is good to merge. Also, in this build Snapshot, Bucket, Autoscaling and EIP sequences were all saving logs as "Build artifacts". (I am working on including screenshot on fail, have most of the code there ready.)

coyote240:
Retest this, please.

PR #1893, opened 2016-04-12
Gui 2466 several fixes.

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2466

### DIFF
--- a/eucaconsole/static/js/pages/metrics.js
+++ b/eucaconsole/static/js/pages/metrics.js
@@ -267,7 +267,9 @@ angular.module('MetricsPage', ['LandingPage', 'CloudWatchCharts', 'EucaConsoleUt
             });
             $scope.metricForAlarm = Object.assign(metric[0]);
             $scope.metricForAlarm.dimensions = dims;
-            ModalService.openModal('createAlarm');
+            $timeout(function() {
+                ModalService.openModal('createAlarm');
+            });
         };
     })
 ;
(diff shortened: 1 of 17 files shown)


eucaconsole/static/js/pages/metrics.js
```diff
@@ -255,8 +255,13 @@ angular.module('MetricsPage', ['LandingPage', 'CloudWatchCharts', 'EucaConsoleUt
             if (!Array.isArray(metric)) {
                 metric = [metric];
             }
+            names = "";
             metric.forEach(function(row) {
                 row.resources.forEach(function(res) {
```
coyote240: Kind of a golden rule for javascript is to never += a string in a loop.  Big memory churn, just like C.  Instead, push values onto a list and join at the end.

eucaconsole/static/js/pages/metrics.js
```diff
@@ -265,9 +270,13 @@ angular.module('MetricsPage', ['LandingPage', 'CloudWatchCharts', 'EucaConsoleUt
                     }
                 });
             });
+            names = names.join('');
             $scope.metricForAlarm = Object.assign(metric[0]);
             $scope.metricForAlarm.dimensions = dims;
-            ModalService.openModal('createAlarm');
+            $scope.metricForAlarm.names = names;
+            $timeout(function() {
```
coyote240: Why the timeout?

dkavanagh: This was needed to allow the digest cycle to complete and update the DOM bindings before open happens.

PR #1900, opened 2016-04-19
GUI-2197, GUI-2199, GUI-2539: Better handling of dimension choices for Alarm Details page and Copy Alarm dialog

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2539

Includes various fixes for:
- https://eucalyptus.atlassian.net/browse/GUI-2197
- https://eucalyptus.atlassian.net/browse/GUI-2199
- https://eucalyptus.atlassian.net/browse/GUI-2523
- https://eucalyptus.atlassian.net/browse/GUI-2531
- https://eucalyptus.atlassian.net/browse/GUI-2547
- https://eucalyptus.atlassian.net/browse/GUI-2548


eucaconsole/models/alarms.py
```diff
@@ -24,9 +24,14 @@
...
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import re
 from itertools import chain
+from operator import itemgetter
+
+import simplejson as json
 
 from ..i18n import _
+from ..views import BaseView, TaggedItemView, boto_error_handler
```
dkavanagh: I think importing views into a "models" package is the wrong way around. If this is really extending a base view, put it in views. Can the code here be package better to not require BaseView?

eucaconsole/models/alarms.py
```diff
@@ -94,3 +99,169 @@ def get_resource_alarm_status(cls, resource_id, alarms):
...
+            choices += ebs_choices
+
+        dimension_choices = list(chain.from_iterable(choices))
+
+        if custom_ns and self._none_selected(dimension_choices):
+            dimension_choices.append({'label': _('Select dimension...'), 'value': '', 'selected': True})
+
+        return dimension_choices
+
+    def _get_instance_choices(self):
```
dkavanagh: Don't all these methods duplicate stuff already in forms/__init__.py ?

eucaconsole/models/alarms.py
```diff
@@ -24,9 +24,15 @@
...
 
+import re
 from itertools import chain
+from operator import itemgetter
 
+import simplejson as json
+
+from ..constants.instances import AWS_INSTANCE_TYPE_CHOICES
 from ..i18n import _
+from ..views import BaseView, TaggedItemView, boto_error_handler
```
dkavanagh: IMHO, we should not be importing views into the models package! If this really depends on BaseView, this should be in views. Could this be packaged better to avoid depending on BaseView?

eucaconsole/static/js/pages/alarms/alarm_detail.js
```diff
@@ -108,27 +84,30 @@ angular.module('AlarmDetailPage', [
...
+                    metric: $scope.metric,
+                    dimensions: JSON.stringify(dimensions),
+                    namespace: $scope.namespace,
+                    duration: $scope.duration,
+                    statistic: $scope.statistic,
+                    unit: $scope.unit
+                }).then(function(oData) {
+                    var results = oData ? oData.results : '';
+                    var chart = ChartService.renderChart($scope.target, results, {
+                        unit: oData.unit || scope.unit
```
dkavanagh: Both the old and new versions do something different and neither version actually produced a graph with data. I commented on the GUI-2197 ticket with more info.

eucaconsole/static/js/widgets/create-alarm-modal/create-alarm.js
```diff
@@ -94,14 +98,18 @@ angular.module('CreateAlarmModal', [
                 if (!$scope.namespace.match(',')) {  // Avoid breaking namespace when multiple NS are passed to directive
                     $scope.alarm.metric.namespace = $scope.namespace;
                 }
-                $scope.alarm.dimensions = $scope.dimensions;
+                $scope.alarm.dimensions = attrs.alarmName ? JSON.stringify($scope.dimensions) : $scope.dimensions;
```
dkavanagh: Could you explain when dimensions should be set to alarmName?

dkavanagh: Is testing for alarm name (which indicates copy?) sufficient, or something more concrete, like the type of $scope.dimensions?

dkavanagh: seeing it used below... hmm.. maybe as good as anything here.

eucaconsole/views/__init__.py
```diff
@@ -532,14 +532,20 @@ def serialize_tags(tags):
         return BaseView.escape_json(json.dumps(serialized_tags))
 
     @staticmethod
-    def get_display_name(resource, escapebraces=True):
+    def get_display_name(resource, escapebraces=True, id_first=False):
```
dkavanagh: Where do we deviate from the standard?

eucaconsole/views/alarms.py
```diff
@@ -419,9 +420,9 @@ def cloudwatch_alarm_view(self):
 
         self.render_dict.update(
             metric_display_name=METRIC_TITLE_MAPPING.get(self.alarm.metric, self.alarm.metric),
-            dimensions=dimensions,
             alarm_actions_json=json.dumps(alarm_actions),
-            dimension_options=dimension_options
+            dimension_options=dimension_options,
+            dimension_options_json=json.dumps(dimension_options),
```
dkavanagh: looks like we sent dimensions in 2 ways before and now, why?

eucaconsole/models/alarms.py
kamalgill: @dkavanagh: Agreed... a prior version of the Dimension class inherited from object, but I had forgot to wrap the API calls in boto_error_handler.  I'll move this to the views.alarms module

eucaconsole/models/alarms.py
kamalgill: @dkavanagh: Yeah, I forgot about the ChoicesManager class here... will refactor the choices to leverage the methods there.

eucaconsole/views/__init__.py
kamalgill: We needed to deviate from the standard when listing dimension choices. The default display name uses the format `name (ID)`, but dimensions are better read as `resourceType = ID (name)` (e.g. `InstanceId = i-123456 (my-instance)`

eucaconsole/views/alarms.py
kamalgill: @dkavanagh: dimension_options is for the alarm details page, while the json version is for choices in the Copy Alarm dialog on that page

dkavanagh: That's a trivial conversion, client-side. Why not do that vs sending both across the wire?

dkavanagh: didn't realize it went right into template.. nbd.

eucaconsole/static/js/pages/alarms/alarm_detail.js
kamalgill: @dkavanagh : Good catch... will troubleshoot in this branch/PR

eucaconsole/templates/cloudwatch/alarms_detail.pt
```diff
@@ -183,6 +184,21 @@
...
+                    <span>Are you sure you want to delete alarm</span>
+                    <strong>{{ alarm.name }}</strong><span i18n:translate="">?</span>
+                </p>
+                <div class="row buttons">
+                    <button id="delete-alarm-btn" class="button expand" ng-click="deleteAlarm($event)" i18n:translate="">
+                        <span i18n:translate="">Yes, Delete Alarm</span>
+                    </button>
+                </div>
+            </div>
+            <!--! /end modal dialogs -->
```
dkavanagh: I was going to merge this, but realize you'll get dinged on this dialog not matching the wireframe. https://eucalyptus.atlassian.net/wiki/display/3UD/Delete+alarms
There are already 2 copies of this dialog, the one in dialogs and the one in the alarm-state-view: https://github.com/eucalyptus/eucaconsole/blob/master/eucaconsole/templates/cloudwatch/alarm_state_view.pt#L35-L66

kamalgill: @dkavanagh I copied the existing dialog and trimmed the unneeded conditions to get it down to this, which was just easier to inline as the dimensions are already on the alarm details page and probably don't need to be repeated in the dialog.  I'll wait for Jenny to comment on GUI-2548, which is in UX review atm.

dkavanagh: I just went through this w/ Jenny, so I was trying to help you.

PR #1902, opened 2016-04-20
GUI-2537 quick solution of disabling 'create alarm' when chart has mo…

dkavanagh:
…re than one series

https://eucalyptus.atlassian.net/browse/GUI-2537

### DIFF
--- a/eucaconsole/static/sass/eucaconsole.scss
+++ b/eucaconsole/static/sass/eucaconsole.scss
@@ -1337,17 +1337,14 @@ $btn-group-radius: 10px;
 }
 
 /* -----------------------------------------
-   links
+   Disabled drop-down items
 ----------------------------------------- */
-a {
-    &[disabled] {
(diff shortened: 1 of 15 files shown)


eucaconsole/static/sass/eucaconsole.scss
```diff
@@ -1337,6 +1337,16 @@ $btn-group-radius: 10px;
 }
 
 /* -----------------------------------------
+   links
+----------------------------------------- */
+a {
+    &[disabled] {
+        // important here to override f-dropdown style in disabled case (using ng-disabled)
+        color: $hpe-lightgray !important;
```
kamalgill: @dkavanagh : We should add `pointer-events: none` to this selector to make the link truly disabled instead of just appearing disabled, right?

dkavanagh: Sure, didn't know we could do that. I ensured code doesn't run in listener instead.

kamalgill:
Looks good to merge

PR #1909, opened 2016-04-22
GUI-2548, GUI-2521: Reuse delete alarm confirmation dialog on alarm details page

kamalgill:
Includes fix for chrome bold/strong tag rendering issue (GUI-2521)

Fixes:
- https://eucalyptus.atlassian.net/browse/GUI-2548
- https://eucalyptus.atlassian.net/browse/GUI-2521

### DIFF
--- a/eucaconsole/static/sass/pages/alarm-detail.scss
+++ b/eucaconsole/static/sass/pages/alarm-detail.scss
@@ -46,6 +46,26 @@ form.alarm-update {
     }
 }
 
-#delete-alarm-btn {
-    margin-top: 1rem;
+#delete-alarm-modal {
+    p {
+        margin: 0 0 7px;
+    }
+    .header {
+        color: #444;
+        text-transform: uppercase;
+        font-size: $euca-font-size;
+        font-weight: bold;
+        .columns {
+            padding: 0;
+        }
+    }
+    .alarm-dimensions {
+        margin: 0 0 20px;
+        .columns {
+            padding: 0;
+        }
+    }
+    .row.control {
+        margin: 0 0 15px;
+    }
 }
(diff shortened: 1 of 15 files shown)


eucaconsole/static/sass/pages/alarm-detail.scss
```diff
@@ -45,3 +45,27 @@ form.alarm-update {
...
+    .alarm-dimensions {
+        margin: 0 0 20px;
+        .columns {
+            padding: 0;
+        }
+    }
+    .row.control {
+        margin: 0 0 15px;
+    }
+}
```
dkavanagh: why not have this in common scss file which gets included wherever dialog used?

kamalgill: @dkavanagh: What common scss file are you referring to?  This is pretty specific to teh delete alarm modal, and these styles were copied from the ones @coyote240 created for the dialog on the alarms landing page (see alarms.scss).

kamalgill: @dkavanagh Ah, I see what you're saying... I'll put these in a common scss file to be used across the three places where the delete confirmation dialog is needed.  BTW, I noticed the delete alarm from the alarms list on the resource monitoring pages doesn't reuse the existing delete alarm dialog.  Is that intentional?

dkavanagh: I thought hard about including the modal in that widget. Since that widget is used in 4 pages, I hesitated adding more dependencies. I stripped down the DOM for the one already in dialogs and put that into the remove dialog I already had. Would be happy to re-use existing scss though if it were in a common file I could include.

kamalgill: @dkavanagh : Common scss file created in https://github.com/eucalyptus/eucaconsole/commit/26d31f335083992e5e99a27263de56e7f49a764c .  Also added that to alarm-state-view.scss, so feel free to have a look there as well?

PR #1912, opened 2016-04-25
GUI-2559: Handle alarm actions updates on Alarm Details page

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2559


dkavanagh:
I want to be sure I understand the big change here. Was the alarm action editor previously pushing action changes itself, and now the updates are handled by modifying alarm actions in the alarm itself and saving the alarm as a whole?

kamalgill:
@dkavanagh : No, the problem here was the alarm details page didn't include the code to handle alarm actions updates when GUI-2197 was implemented.  The alarm actions editor doesn't push action changes via XHR calls (thank goodness).  Changes to the alarm actions are saved to a local alarmActions scope variable -- that variable and the `$scope.$on('actionsUpdated' ...)` handler wasn't included in alarm-details.js

dkavanagh:
I saw that the route for cloudwatch_alarm_actions was removed and associated client/server code.

PR #1913, opened 2016-04-25
GUI-2523: Handle when resource in dimensions is no longer available

kamalgill:
Minor follow-up fix for https://eucalyptus.atlassian.net/browse/GUI-2523 , which was mostly fixed in #1900


eucaconsole/static/js/pages/alarms/alarm_detail.js
```diff
@@ -12,6 +12,10 @@ angular.module('AlarmDetailPage', [
             scope.alarmDimensions = scope.alarm.dimensions;  // Leveraged in delete alarm confirmation dialog
             // Need stringified form on details page (and Copy Alarm dialog) to set current dimension choice
             scope.alarm.dimensions = JSON.stringify(scope.alarm.dimensions);
+            if (parseInt(attrs.invalidDimensions, 10) || 0) {
+                // Handle when resource in dimensions is no longer available
+                scope.alarm.dimensions = '';
+            }
```
dkavanagh: While it is a rarer case, this isn't as nice when there are 2 dimensions and 1 is no longer there.

PR #1915, opened 2016-04-26
GUI-2543: Rework scaling policies page to better handle multi-alarm policies

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2543

### DIFF
--- a/eucaconsole/views/scalinggroups.py
+++ b/eucaconsole/views/scalinggroups.py
@@ -29,6 +29,8 @@ Pyramid views for Eucalyptus and AWS scaling groups
 
 """
 from dateutil import parser
+from operator import itemgetter
+
 import simplejson as json
 import time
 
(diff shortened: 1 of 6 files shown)


eucaconsole/views/scalinggroups.py
```diff
@@ -700,13 +699,34 @@ def __init__(self, request):
...
+            if hasattr(policy, 'alarms'):
+                for alarm in self.alarms:
+                    for policy_alarm in policy.alarms:
+                        if alarm.name == policy_alarm.name and alarm not in policy_alarms:
+                            policy_alarms.append(alarm)
+            encoded_policy_name = policy.name.encode('UTF-8')
+            policy_ids[encoded_policy_name] = md5(encoded_policy_name).hexdigest()[:8]
+            scale_text_prefix = _('Remove') if policy.scaling_adjustment < 0 else _('Add')
+            scale_type = _('instances') if policy.adjustment_type == 'ChangeInCapacity' else '%'
+            if policy.adjustment_type == 'ChangeInCapacity' and abs(policy.scaling_adjustment) == 1:
```
dkavanagh: Would 1 really be the only value here? Do we need > 0 instead?

kamalgill: @dkavanagh: The check here is for the 'instance' vs 'instances' label

dkavanagh: oic.. good to merge then.

PR #1916, opened 2016-04-26
GUI-2564: Ensure proper instance count is reflected per zone on ELB Instances page

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2564


kamalgill:
Makes changes to ELB instance selector to account for changes in https://github.com/eucalyptus/eucaconsole/commit/fc3a68086b6527333a02537f0db60e1e08488f51#diff-87ccf491f864dfb6e251acc2b973bf29R434

PR #1921, opened 2016-04-28
Gui 2561 fixed namespace, duplicate js import and unique_name

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2561
https://eucalyptus.atlassian.net/browse/GUI-2560

### DIFF
--- a/eucaconsole/views/metrics.py
+++ b/eucaconsole/views/metrics.py
@@ -300,7 +300,7 @@ class CloudWatchMetricsJsonView(BaseView):
                     cat_metrics.append(dict(
                         cat_name=cat['name'],
                         namespace=metric[2],
-                        unique_id=metric_name + '-' + '-'.join([dim[1][0] for dim in metric_dims]),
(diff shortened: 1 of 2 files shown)


eucaconsole/views/metrics.py
```diff
@@ -278,19 +278,18 @@ def cloudwatch_metrics_json(self):
                     } for item in items if item.namespace not in STD_NAMESPACES and item.dimensions]
                 if cat['name'] in ['ec2allinstances', 'elball']:
                     tmp = set([met['name'] for met in tmp])
+                    tmp = [([], met, item.namespace) for met in tmp]
```
kamalgill: @dkavanagh : PyCharm indicates the following error for this line... "Local variable 'item' might be referenced before assignment".  Where is item coming from here?

dkavanagh: I don't know how that worked, yet it was finding a good namespace when I checked. Let me make that more deterministic.

PR #1929, opened 2016-05-06
GUI-2354: Incorporate alarm thresholds in charts on alarm details page and in create/copy alarm dialog

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2354


eucaconsole/static/js/widgets/create-alarm-modal/create-alarm.js
```diff
@@ -351,4 +356,117 @@ angular.module('CreateAlarmModal', [
 
         }
     };
-});
+})
+.directive('metricChart', function () {
```
coyote240: Why did this directive move?

eucaconsole/static/js/widgets/create-alarm-modal/create-alarm.js
```diff
@@ -351,4 +356,117 @@ angular.module('CreateAlarmModal', [
...
+        restrict: 'A',
+        scope: {
+            metric: '@',
+            namespace: '@',
+            duration: '=',
+            statistic: '=',
+            unit: '@',
+            dimensions: '=',
+            threshold: '@',
+            formName: '@formname'
```
coyote240: Naming the attribute form-name will translate to formName, if you want.

eucaconsole/static/js/widgets/create-alarm-modal/create-alarm.js
kamalgill: It moved to reuse it across the alarm details page and create alarm dialog

eucaconsole/static/js/widgets/create-alarm-modal/create-alarm.js
kamalgill: Yeah, I realize that, but I'm going for the "explicit is better than implicit" approach here.  No big deal to change it though if you have a strong opinion to do that.

coyote240: I don't, it's just a note.  Standard angularjs naming.

eucaconsole/static/js/widgets/create-alarm-modal/create-alarm.js
```diff
@@ -351,4 +356,117 @@ angular.module('CreateAlarmModal', [
...
+                    return;
+                }
+                var parsedDims = angular.isObject(newVal) ? newVal : JSON.parse(newVal);
+                var resourceLabel = '';
+                var resourceLabels = [];
+                var dimensionField = angular.element('form[name="' + $scope.formName + '"]').find('[name="dimensions"]');
+                var selectedDimField = dimensionField.find('[selected]');
+                if (selectedDimField.length && newVal === $scope.dimensions) {
+                    resourceLabel = selectedDimField.text();
+                }
```
coyote240: This is pretty heavy reliance on DOM traversal, which is not encouraged, least of all in a controller.  Also, particularly on line 422, is a bit opaque.  Can you not refer to a value on the $scope instead?

kamalgill: @coyote240 : $scope.dimensions would give us the value, but we need the label here for the chart legend when the selected dimensions are changed.  We could build a value/label mapping for the dimensions choices and look up the label there, but we'd need to pass those to the metricChart directive and I'd like to avoid that overhead.

coyote240: DOM traversal is overhead.  Opaque, difficult to read code is overhead.  Both, I see as greater than shifting towards using bindings.

PR #1930, opened 2016-05-09
GUI-2582: Fix CF stack status colors on details page Events tab

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2582


dkavanagh:
This breaks magic search because (as I said in the ticket), the status you capitalized is not used on the client side, rather for filtering search results.

PR #1934, opened 2016-05-11
GUI-2478, Adam's attempt, interactive overlay

coyote240:


eucaconsole/static/js/pages/cloudwatch_charts.js
```diff
@@ -30,6 +30,59 @@ angular.module('CloudWatchCharts', ['EucaConsoleUtils', 'ChartAPIModule', 'Chart
...
+                        mouseY: y
+                    });
+                    graph.interactiveLayer.renderGuideLine(null);
+                    return;
+                }
+
+                var pointXValue = graph.interactiveLayer.xScale().invert(x);
+                graph.interactiveLayer.dispatch.elementMousemove({
+                    mouseX: x,
+                    mouseY: (y % 362) - (margin.bottom + margin.top),
```
dkavanagh: I will add that the chart height is driven by a css rule, but obviously there are titles and other stuff.
    svg {
        height: 300px !important;
    }

Not sure if there's a nicer way to drive 362, or if this is good enough for now.

coyote240: I started out getting the height of the chart container, but as you point out, there are titles and other stuff.  I don't like this, but it could get pretty wicked to derive the height from a container.
That being said, I may be able to put a directive on the div.chart-wrapper element and see if it can report it's own height.  Think it's worth it for now?

PR #1945, opened 2016-05-16
Gui 2578 fix region switching functionality.

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2578


eucaconsole/models/auth.py
```diff
@@ -240,7 +240,7 @@ def euca_connection(ufshost, port, region, access_id, secret_key, token, conn_ty
                 ufshost, port, 'euca', access_id, secret_key, token, 'ec2', dns_enabled
             )
             regions = RegionCache(conn).get_regions(ufshost)
-            region = [region.endpoint for region in regions if region.name == region]
+            region = [reg.endpoint for reg in regions if reg.name == region]
```
kamalgill: It's confusing to name a list as a singular variable. A better name for this would be 'endpoints', or perhaps 'region_endpoints'.

dkavanagh: I see what you're saying, but this actually narrows down to a single endpoint. which is extracted 2 lines down.

kamalgill: These line would be more clear if written as follows...

```
    endpoints = [region.endpoint for region in regions if region.name == region]
    if endpoints:
        endpoint = endpoints[0]
```

dkavanagh: I guess it begs the question, should a list of 1 be pluralized. It would certainly be more accurate to refer to it as an endpoint vs region though.

eucaconsole/layout.py
```diff
@@ -98,7 +98,9 @@ def __init__(self, context, request):
                     self.has_regions = False
         if hasattr(self, 'regions'):
             self.selected_region = self.request.session.get('region', self.default_region)
-            if self.selected_region == '' or self.selected_region == 'undefined':
+            if (self.selected_region == '' or
+                self.selected_region == 'undefined' or
+                self.selected_region == 'euca'):
```
coyote240: Is 'undefined' here because the JavaScript value 'undefined' is making its way to the server as a string?

dkavanagh: I believe so. It was in December when I added that check, so not 100% sure.

eucaconsole/models/auth.py
```diff
@@ -240,9 +240,10 @@ def euca_connection(ufshost, port, region, access_id, secret_key, token, conn_ty
                 ufshost, port, 'euca', access_id, secret_key, token, 'ec2', dns_enabled
             )
             regions = RegionCache(conn).get_regions(ufshost)
-            region = [region.endpoint for region in regions if region.name == region]
-            if region:
-                endpoint = region[0]
+            # result will only ever be 1 or none
+            endpoint = [reg.endpoint for reg in regions if reg.name == region]
+            if endpoint:
```
coyote240: Not a big deal, but I wonder if a reduce better expresses what's happening with the regions here, with the zero index below.  If the comprehension does in fact return an empty list, the zero index will error.

eucaconsole/layout.py
coyote240: Not worth spending time on, but worth keeping in mind that the client may be leaking into the service :-)

eucaconsole/models/auth.py
dkavanagh: that's why I do the check above that.

PR #1946, opened 2016-05-16
GUI-2551, prevent expression interpolation in values returned from api.

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-2551

### DIFF
--- a/eucaconsole/static/js/pages/alarms/alarm_detail.js
+++ b/eucaconsole/static/js/pages/alarms/alarm_detail.js
@@ -5,28 +5,35 @@ angular.module('AlarmDetailPage', [
 .directive('alarmDetail', ['eucaRoutes', function (eucaRoutes) {
     return {
         restrict: 'A',
-        link: function (scope, element, attrs) {
-            scope.alarm = JSON.parse(attrs.alarmDetail);
-            console.log(scope.alarm);
-            scope.alarm.actions = scope.alarm.actions || [];
-            scope.alarms = [scope.alarm];  // Delete alarm confirmation dialog expects a list of alarms
-            scope.expanded = true;
-            scope.alarmDimensions = scope.alarm.dimensions;  // Leveraged in delete alarm confirmation dialog
-            // Need stringified form on details page (and Copy Alarm dialog) to set current dimension choice
-            scope.alarm.dimensions = JSON.stringify(scope.alarm.dimensions);
+        compile: function (tElement, tAttrs) {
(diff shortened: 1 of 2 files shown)


eucaconsole/static/js/pages/alarms/alarm_detail.js
```diff
@@ -5,27 +5,35 @@ angular.module('AlarmDetailPage', [
...
-        link: function (scope, element, attrs) {
-            scope.alarm = JSON.parse(attrs.alarmDetail);
-            scope.alarm.actions = scope.alarm.actions || [];
-            scope.alarms = [scope.alarm];  // Delete alarm confirmation dialog expects a list of alarms
-            scope.expanded = true;
-            scope.alarmDimensions = scope.alarm.dimensions;  // Leveraged in delete alarm confirmation dialog
-            // Need stringified form on details page (and Copy Alarm dialog) to set current dimension choice
-            scope.alarm.dimensions = JSON.stringify(scope.alarm.dimensions);
+        compile: function (tElement, tAttrs) {
+            var alarm = JSON.parse(tAttrs.alarmDetail);
```
coyote240: Seems that deserializing the alarm details before the link function preempts expression interpolation.

eucaconsole/static/js/pages/alarms/alarm_detail.js
```diff
@@ -52,23 +60,6 @@ angular.module('AlarmDetailPage', [
                     });
             };
 
-            $scope.$on('actionsUpdated', function (event, actions) {
```
coyote240: This seems to have come over in a cut'n'paste.  Actions are collated in a watch below.

eucaconsole/static/js/pages/alarms/alarm_detail.js
dkavanagh: So, the scope.alarm value gets bound to the UI field, but not evaluated? I believe what you're saying, but wanted to understand what's happening more.

coyote240: As I understand it.  Now, notice that it's not scope.alarm that is bound to the template but scope.alarm.description.  It's getting pretty deep into angularjs internals that aren't documented super-great, but it seems that by deserializing and assigning to scope.alarm before linking we get ahead of expression evaluation.  I wouldn't do this in too many places, but here there is certainly no desire to allow expressions to be stored or evaluated.

PR #1948, opened 2016-05-17
GUI-2603, update button scss mixin, tighten submit strictures.

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-2603

### DIFF
--- a/eucaconsole/static/js/pages/custom_filters.js
+++ b/eucaconsole/static/js/pages/custom_filters.js
@@ -18,7 +18,7 @@ angular.module('CustomFilters', [])
 })
 .filter('ellipsis', function () {
     return function (line, num) {
-        if( line === null || line.length === 0 ){
+        if(!line) {
             return "";
         }else if( line.length <= num ){
             return line;
--- a/eucaconsole/static/js/widgets/tag-editor/tag-editor.js
+++ b/eucaconsole/static/js/widgets/tag-editor/tag-editor.js
@@ -13,6 +13,9 @@ angular.module('TagEditorModule', ['EucaConsoleUtils'])
                 return attributes.template;
             },
             controller: ['$scope', '$window', function ($scope, $window) {
+                $scope.newTagKey = '';
+                $scope.newTagValue = '';
+
                 $scope.addTag = function ($event) {
                     $event.preventDefault();
 
@@ -77,7 +80,7 @@ angular.module('TagEditorModule', ['EucaConsoleUtils'])
         };
     }])
     .directive('tagName', function () {
-        var validPattern = /^(?!aws:)(?!euca:).{0,128}$/;
+        var validPattern = /^(?!aws:)(?!euca:).{1,128}$/;
         return {
             require: 'ngModel',
             link: function (scope, element, attrs, ctrl) {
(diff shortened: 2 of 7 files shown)


eucaconsole/static/js/pages/custom_filters.js
```diff
@@ -18,7 +18,7 @@ angular.module('CustomFilters', [])
 })
 .filter('ellipsis', function () {
     return function (line, num) {
-        if( line === null || line.length === 0 ){
+        if(!line) {
```
coyote240: Missing that one, critical check, is it undefined?  All three are captures in truthiness.

coyote240:
retest this please

kamalgill:
Looks good to me, passing to @dkavanagh for another stamp of approval...

eucaconsole/static/js/widgets/tag-editor/tag-editor.js
```diff
@@ -86,7 +91,7 @@ angular.module('TagEditorModule', ['EucaConsoleUtils'])
         };
     })
     .directive('tagValue', function () {
-        var validPattern = /^(?!aws:).{0,256}$/;
+        var validPattern = /^(?!aws:).{1,256}$/;
```
dkavanagh: Are we handling the empty tag value here? In that case, 0 lower end would be OK.

kamalgill: We don't support tags with empty values yet.  See comments in https://eucalyptus.atlassian.net/browse/GUI-2603

eucaconsole/templates/tag-editor/tag-editor.pt
```diff
@@ -39,18 +39,17 @@
                         Tag names must be unique
                     </span>
                 </div>
-                <div id="autoscale-tag-value-input-div" class="large-8 medium-7 small-5 columns">
+                <div id="autoscale-tag-value-input-div" class="large-6 medium-6 small-4 columns">
```
dkavanagh: I'll point out a feature of this grid system that if in all views, the number of columns is the same, you only need to specify the small. In this case, you could get away with "medium-6 small-4" and get the same results.

dkavanagh:
retest this please

aliceh:
Retest this please

aliceh:
Retest this please

aliceh:
@kamalgill  @coyote240 In this branch, when desired capacity, min or max of autoscaling group is changed the "Save changes" button does not become active. That is the reason for the consitent failures in  autoscaling sequence.

eucaconsole/static/js/widgets/tag-editor/tag-editor.js
```diff
@@ -75,7 +80,7 @@ angular.module('TagEditorModule', ['EucaConsoleUtils'])
         };
     }])
     .directive('tagName', function () {
-        var validPattern = /^(?!aws:)(?!euca:).{0,128}$/;
+        var validPattern = /^(?!aws:)(?!euca:).{1,128}$/;
```
dkavanagh: Requiring values in this field (and the value field) causes the form $invalid flag to be true. That disables the SaveChanges button when it should be enabled.

aliceh:
Retest this please

dkavanagh:
@aliceh nothing was fixed. Not worth retesting till Adam gets a chance to look at this tomorrow.

dkavanagh:
Changes make sense. Allowing 0 size for field level validation, but disallowing add tag if values aren't populated.

PR #1950, opened 2016-05-19
GUI-2158: Workaround to avoid resetting selected ELB security policy options …

kamalgill:
… when policy dialog is subsequently opened for HTTPS listeners in the Create ELB wizard

Fixes https://eucalyptus.atlassian.net/browse/GUI-2158

### DIFF
--- a/eucaconsole/static/js/widgets/elb_security_policy_editor.js
+++ b/eucaconsole/static/js/widgets/elb_security_policy_editor.js
@@ -25,6 +25,22 @@ angular.module('ELBSecurityPolicyEditor', ['EucaConsoleUtils'])
             $('#ssl_protocols').chosen({width: '100%'});
             $('#ssl_ciphers').chosen({width: '100%', search_contains: true});
         };
+        $scope.policyModal.on('opened.fndtn.reveal', function () {
+            // TODO: Find a better workaround here
+            // Although this isn't ideal, we're re-jiggering the field values when the ELB security policy
+            //   modal is subsequently opened in the Create ELB wizard to display the selected policy settings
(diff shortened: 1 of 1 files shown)


eucaconsole/static/js/widgets/elb_security_policy_editor.js
```diff
@@ -25,6 +25,22 @@ angular.module('ELBSecurityPolicyEditor', ['EucaConsoleUtils'])
...
+                if ($scope.predefinedPolicy) {
+                    $scope.policyModal.find('#predefined_policy').val($scope.predefinedPolicy);
+                }
+            } else {  // Custom security policy
+                $scope.policyModal.find('#policy-type-new').prop('checked', true);
+                $scope.policyModal.find('#ssl_protocols').val($scope.sslProtocols).trigger('chosen:updated');
+                $scope.policyModal.find('#ssl_ciphers').val($scope.sslCiphers).trigger('chosen:updated');
+                $scope.policyModal.find('#server_order_preference').prop('checked', $scope.sslServerOrderPref);
+            }
+        });
```
dkavanagh: It would have been nice to rely on angular here. Ideally, chose from one of the chosen directives and then set the model and you're done.

coyote240: This would be a pretty straight-forward rewrite as a directive.  As it is, the reliance on jquery is so heavy, I don't see an alternative as most of the code in this controller is work-arounds for not using model binding in the first place.
If we're worried about time, I'm inclined to pass this and have @kamalgill enter a ticket for a rewrite.

coyote240:
retest this, please

coyote240:
@kamalgill Maybe take a look and see why CI is failing?

PR #1952, opened 2016-05-19
GUI-2449: Allow configuration of scaling group termination policies in create scaling group wizard

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2449

### DIFF
--- a/eucaconsole/forms/scalinggroups.py
+++ b/eucaconsole/forms/scalinggroups.py
@@ -98,6 +98,14 @@ class BaseScalingGroupForm(BaseSecureForm):
             validators.InputRequired(message=health_check_period_error_msg),
         ],
     )
+    termination_policies_error_msg = _(u'At least one termination policy is required')
+    termination_policies_help_text = _(u'Add termination policies in the order they should be executed.')
+    termination_policies = wtforms.SelectMultipleField(
+        label=_(u'Termination policies'),
+        validators=[
+            validators.InputRequired(message=termination_policies_error_msg),
+        ],
+    )
 
     def __init__(self, request, scaling_group=None, launch_configs=None,
                  autoscale_conn=None, ec2_conn=None, vpc_conn=None, elb_conn=None, **kwargs):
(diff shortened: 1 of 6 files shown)


eucaconsole/templates/scalinggroups/scalinggroup_wizard.pt
```diff
@@ -71,6 +71,10 @@
                                     ${panel('form_field', field=create_form['max_size'], leftcol_width_large=6, rightcol_width_large=6, maxlength=2, ng_attrs={'model': 'maxSize', 'change': 'handleSizeChange()'})}
                                 </div>
                             </div>
+                            <div>&nbsp;</div>
```
coyote240: What's the empty div for?

eucaconsole/forms/scalinggroups.py
```diff
@@ -132,9 +141,11 @@ def __init__(self, request, scaling_group=None, launch_configs=None,
         self.max_size.error_msg = self.max_size_error_msg
         self.min_size.error_msg = self.min_size_error_msg
         self.health_check_period.error_msg = self.health_check_period_error_msg
+        self.termination_policies.error_msg = self.termination_policies_error_msg
```
dkavanagh: Is this redundant, given the validator already has the error message?

dkavanagh: What I was asking about is that the same error message is pulled into the field validator and now being set on the field itself. Does it need to be in both places, or does the validator override any field level error message?

kamalgill: This is our usual WTForms way of surfacing the error message to the template (see the form_field_row panel and form_field_row.pt).

dkavanagh:
retest this please

eucaconsole/forms/scalinggroups.py
dkavanagh: So, the difference is the validator message is used client-side where the other is returned if server-side validation fails?

kamalgill: Correct.  When server-side validation fails, WTForms returns a list of validation messages.  We don't leverage that in many places, since Foundation handles the validation client-side before ever reaching server-side validation.  The message in `validators.InputRequired(message=termination_policies_error_msg),` is probably unnecessary here, but there may be cases where server-side validation is reached so I'm inclined to leave it.

dkavanagh:
Looks like that volume test failed again. Going to merge since this is unrelated to volumes.

PR #1954, opened 2016-05-19
GUI-2614 fixed facet and text filtering. fixed history reload and breadcrumbs

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2614


dkavanagh:
retest this please

coyote240:
Merging, but food for thought:  If the inputs are stateful, does the result-set need to be stateful?  I'll bet we can make the ng-repeat the result of the filter function and spare the state variables.

dkavanagh:
@coyote240 Are we talking about having a filter in the ng-repeat that does the filtering that's done between the various lists now? We could, but at ng-conf, that was replaced with doing more or less "filter when the data changes, not the view".

PR #1957, opened 2016-05-19
GUI-2274 narrowed comparison so EIPAssociation isn't caught in EIP test

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2274

### DIFF
--- a/eucaconsole/views/stacks.py
+++ b/eucaconsole/views/stacks.py
@@ -386,7 +386,7 @@ class StackStateView(BaseView):
         elif "AWS::EC2::" in res_type:
             if "SecurityGroup" in res_type:
                 url = request.route_path('securitygroup_view', id=resource_id)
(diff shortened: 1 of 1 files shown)


eucaconsole/views/stacks.py
```diff
@@ -386,7 +386,7 @@ def get_url_for_resource(request, res_type, resource_id):
         elif "AWS::EC2::" in res_type:
             if "SecurityGroup" in res_type:
                 url = request.route_path('securitygroup_view', id=resource_id)
-            elif "EIP" in res_type:
+            elif res_type[10:] == "EIP":
```
coyote240: or, perhaps, res_type[-3:]?

dkavanagh: because "AWS::EC2::" is 10 characters, this felt right.

PR #1981, opened 2016-05-26
GUI-2627: Bump up modal size and bump down font size for alarm history details dialog

kamalgill:
… to avoid cutting off Download button.

Fixes https://eucalyptus.atlassian.net/browse/GUI-2627

### DIFF
--- a/eucaconsole/static/sass/pages/alarm_history.scss
+++ b/eucaconsole/static/sass/pages/alarm_history.scss
@@ -11,6 +11,7 @@ table {
 .alarm-history-details-modal {
     pre {
         margin: 1rem 0 0;
+        font-size: 0.9rem;
     }
 
     .button {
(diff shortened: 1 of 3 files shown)


eucaconsole/static/sass/pages/alarm_history.scss
```diff
@@ -11,6 +11,9 @@ table {
 .alarm-history-details-modal {
     pre {
         margin: 1rem 0 0;
+        font-size: 0.9rem;
+        height: 60vh;
```
coyote240: Seems unnecessarily short.  I tested this with 100vh in FF and Safari and it did a great job.

kamalgill: @coyote240: 60vh is 60% of viewport height, which seems reasonable.  Setting the content area to 100vh (100% of viewport height) explodes the dialog well beyond the viewport.

kamalgill: @coyote240: To provide more context on the 60vh setting, changing that to 100vh explodes the dialog outside the viewport when tested on my external 24-in (1920 x 1200) display.  The dialog doesn't explode when testing on a 15" retina Macbook though.  We could probably bump it up to 75vh, which seems like a good compromise.  Thoughts?

coyote240: @kamalgill That sounds fine.

PR #2007, opened 2016-06-16
GUI-2658 replace object.assign with loop

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2658

### DIFF
--- a/eucaconsole/static/js/pages/metrics.js
+++ b/eucaconsole/static/js/pages/metrics.js
@@ -297,9 +297,7 @@ angular.module('MetricsPage', ['LandingPage', 'CloudWatchCharts', 'EucaConsoleUt
             $scope.metricForAlarm = {};
             // core piece of Object.assign polyfill to replace missing call in IE11
             for (var key in metric[0]) {
-              if (Object.prototype.hasOwnProperty.call(metric[0], key)) {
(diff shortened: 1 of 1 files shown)


eucaconsole/static/js/pages/metrics.js
```diff
@@ -294,7 +294,13 @@ angular.module('MetricsPage', ['LandingPage', 'CloudWatchCharts', 'EucaConsoleUt
                 });
             });
             names = names.join('');
-            $scope.metricForAlarm = Object.assign({}, metric[0]);
+            $scope.metricForAlarm = {};
+            // core piece of Object.assign polyfill to replace missing call in IE11
+            for (var key in metric[0]) {
+              if (Object.prototype.hasOwnProperty.call(metric[0], key)) {
```
coyote240: 'in' works here, rather than hasOwnProperty.

coyote240: Actually, why test for keys on an object who's key's you're iterating over?  Unless there is a risk the object will change during this iteration?

dkavanagh: Basically, I took the working part of the polyfill verbatim to minimize chance for error this late in the release cycle.

PR #2011, opened 2016-06-22
GUI-2675 switch to defining worker_class in startup script

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2675


dkavanagh:
looks like unrelated autoscaling failure.

PR #2017, opened 2016-06-27
GUI-2692, set flag on tag update event.

coyote240:


aliceh:
@coyote240 Autoscaling test successfully completed, Jenkins shows failure because it was unable to extract logs. Good to merge.

PR #2026, opened 2016-07-07
GUI-2704 removed gevent refs since we no longer use it

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2704


aliceh:
Good to merge. Jenkins failed to collect logs in instance test and threw a "Fail"

PR #2037, opened 2016-07-22
GUI-2691: Populate availability zone choices when zone is a CF template parameter

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2691 (for 4.3.1)

### DIFF
--- a/eucaconsole/views/stacks.py
+++ b/eucaconsole/views/stacks.py
@@ -797,7 +797,7 @@ class StackWizardView(BaseView, StackMixin):
 
     def get_availability_zone_options(self):
         conn = self.get_connection()
-        zones = ChoicesManager(conn).availability_zones(self.cloud_type)
+        zones = ChoicesManager(conn).availability_zones(self.cloud_type, add_blank=False)
         return zones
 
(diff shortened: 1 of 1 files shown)


aliceh:
Tests passed, but Jenkins failed to archive artifacts in s.group test. Good to merge.

dkavanagh:
looks good to me

eucaconsole/views/stacks.py
```diff
@@ -793,6 +795,11 @@ def get_vmtype_options(self):
         vmtypes = ChoicesManager(conn).instance_types(self.cloud_type)
         return vmtypes
 
+    def get_availability_zone_options(self):
+        conn = self.get_connection()
+        zones = ChoicesManager(conn).availability_zones(self.cloud_type, add_blank=False)
```
dkavanagh: I'm pondering the add_blank here.  The rest default to True. We've chosen to not provide a default selection here unless the template has a default or there is only 1 value available.

dkavanagh: Now, I'm noticing how other methods to gather values were just using the connection directly and don't have a blank. Yes, setting this to False is correct.

PR #2044, opened 2016-08-12
Aug11 master 5.0 master

dkavanagh:


dkavanagh:
This test an CI w/ centos 7 http://gui-jenkins.qa1.eucalyptus-systems.com:8080/view/Centos%207/job/centos7-user-console-regression-sequence/360/
one failure in instances. @aliceh, is this something we could safely ignore?

PR #2046, opened 2016-08-26
Aug26 master 50master

dkavanagh:


dkavanagh:
The test failures were failure to find newly created bucket on page, or ip addr on landing page. Much of each sequence had already executed, so don't think these are fatal. Going to merge.

PR #2060, opened 2016-09-21
GUI-2736: GUI CORS Configuration Support

kamalgill:
Implements: 
- https://eucalyptus.atlassian.net/browse/GUI-2736
- https://eucalyptus.atlassian.net/browse/GUI-2757

Tested on:
- Chrome 53 on Mac OS 10.10, Windows 10, and Fedora 24
- Firefox 49 on Mac OS 10.10, Windows 10, and Fedora 24
- Safari on Mac OS 10.10
- MS Edge on Windows 10 (with Anniversary Update)

Notes:
- libxml2 and libxslt are required (by python-lxml); see updated dependency notes in README.rst


aliceh:
Retest this please

eucaconsole/static/js/pages/bucket_details.js
```diff
@@ -82,12 +92,92 @@ angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils'])
...
+                    deleteDialog.foundation('reveal', 'close');
+                    $scope.deletingCorsConfig = false;
+                    $scope.hasCorsConfig = false;
+                    Notify.success(response.data.message);
+                }, function error (errData) {
+                    $scope.deleteError = errData.data.message;
+                    $scope.deletingCorsConfig = false;
+                });
+        };
+        $scope.$on('s3:corsConfigSaved', function () {
```
dkavanagh: My understanding is that situations like this are better suited towards shared scope values, not events. When a directive needs to communicate information to the containing scope, just use scope binding. It has far less overhead and is a good practice. Events are better used when the sending and receiving components don't have such a direct connection.

eucaconsole/static/js/thirdparty/codemirror/xml-fold.js
```diff
@@ -0,0 +1,173 @@
+(function() {
```
dkavanagh: is there a minified version?

eucaconsole/static/js/thirdparty/codemirror/closetag.js
```diff
@@ -0,0 +1,93 @@
+/**
```
dkavanagh: Is there a minified version?

eucaconsole/static/js/thirdparty/codemirror/xml.js
```diff
@@ -0,0 +1,332 @@
+CodeMirror.defineMode("xml", function(config, parserConfig) {
```
dkavanagh: Is there a minified version?

eucaconsole/templates/buckets/bucket_details.pt
```diff
@@ -122,11 +123,51 @@
...
         ${panel('bucket_dialogs', landingpage=False, bucket=bucket, versioning_form=versioning_form, delete_form=delete_form)}
         ${panel('create_folder_dialog', bucket_name=bucket_name, create_folder_form=create_folder_form)}
+        <!--! Modal dialog for viewing/editing CORS configuration -->
+        <div class="modal large cors-config-modal" modal="corsConfigModal">
+            <div cors-config-modal=""
+                 template="${request.route_path('render_template', subpath='dialogs/bucket_cors_configuration_dialog')}"
+                 bucket-name="${bucket_name}"
+                 has-cors-config="hasCorsConfig"
+                 cors-config-xml="${cors_configuration_xml}"
+                 sample-cors-config="${sample_cors_configuration}">
```
dkavanagh: Here, is where we'd add the cors-saved, or similar value, mapped to a page scope variable, instead of using events.

eucaconsole/views/buckets.py
```diff
@@ -737,6 +752,41 @@ def bucket_update(self):
             self.request.error_messages = self.details_form.get_errors_list()
         return self.render_dict
 
+    @view_config(route_name='bucket_set_cors_configuration', renderer='json', request_method='PUT', xhr=True)
```
dkavanagh: FWIW, it seems like this and the delete view below really don't depend on much from the view. The init for this view is pretty huge, mostly to support template views. Could these be in an XHR view that is much lighter weight?

eucaconsole/constants/buckets.py
```diff
@@ -0,0 +1,87 @@
...
+# See http://docs.aws.amazon.com/AmazonS3/latest/dev/cors.html
+#
+# NOTE: This schema does not include the "http://s3.amazonaws.com/doc/2006-03-01/" namespace declaration,
+#       so any CORS XML validated against this schema will need namespaces removed
+#
+CORS_XML_RELAXNG_SCHEMA = """
+<element name="CORSConfiguration" xmlns="http://relaxng.org/ns/structure/1.0">
+  <oneOrMore>
+    <element name="CORSRule">
+      <interleave>
```
dkavanagh: Is it a given that these can be in any order? The docs seem to show then in a specific order. If you've tested against AWS with the order changed, that's good enough for me.

eucaconsole/static/js/pages/bucket_details.js
```diff
@@ -82,12 +92,92 @@ angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils'])
...
+                hasCorsConfig: '=',
+                corsConfigXml: '@',
+                sampleCorsConfig: '@'
+            },
+            templateUrl: function (element, attributes) {
+                return attributes.template;
+            },
+            controller: ['$scope', 'CorsService', 'ModalService',
+                function($scope, CorsService, ModalService) {
+                    var pageScope = angular.element('#contentwrap').scope();
```
coyote240: Dangerous at any speed.  Better off adding a directive at a higher level to allow cross-controller communication.

eucaconsole/static/js/pages/bucket_details.js
```diff
@@ -82,12 +92,92 @@ angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils'])
...
+                    }
+                    $scope.setCorsConfiguration = function ($event) {
+                        $event.preventDefault();
+                        $scope.savingCorsConfig = true;
+                        $scope.corsError = '';
+                        var csrfToken = $('#csrf_token').val();
+                        CorsService.setCorsConfig($scope.bucketName, csrfToken, $scope.codeEditor.getValue())
+                            .then(function success (response) {
+                                $scope.savingCorsConfig = false;
+                                pageScope.$broadcast('s3:corsConfigSaved');
```
coyote240: Looks like you're only using pageScope for the broadcast.  Better off using $rootScope instead and eliminating the reference to pageScope entirely.

eucaconsole/static/js/pages/bucket_details.js
coyote240: These components don't have a direct connection.  It's part of the trouble with page controllers.  Now, should we add a directive at the page controller level, the child directive may require the parent directive and have access to it's controller.

eucaconsole/views/buckets.py
```diff
@@ -756,6 +806,15 @@ def bucket_update_versioning(self):
             self.request.error_messages = self.versioning_form.get_errors_list()
         return self.render_dict
 
+    def get_controller_options_json(self):
+        return BaseView.escape_json(json.dumps({
+            'bucket_name': self.bucket.name,
+            'cors_config_xml': self.cors_configuration_xml,
+            'has_cors_config': bool(self.cors_configuration_xml),
```
coyote240: Unneeded state variable here, actually creates overhead maintaining it later in the JS code.  Better off simply evaluating the cors_config_xml client-side for presence.  Get rid of this and you can remove all references to hasCorsConfig client-side, then you can simply check the value of corsConfigXml and set it accordingly.  eg. bucket_details.js:139,173

eucaconsole/static/js/thirdparty/codemirror/xml-fold.js
kamalgill: There doesn't seem to be

eucaconsole/constants/buckets.py
kamalgill: Yes, the config appears to be order-independent for items in a CORSRule element

eucaconsole/routes.py
```diff
@@ -223,6 +223,8 @@ def instances_landing(self):
     Route(name='bucket_update', pattern='/buckets/{name}/update'),
     Route(name='bucket_delete', pattern='/buckets/{name}/delete'),
     Route(name='bucket_delete_keys', pattern='/buckets/{name}/delete_keys'),
+    Route(name='bucket_delete_cors_configuration', pattern='/buckets/{name}/delete_cors_configuration'),
+    Route(name='bucket_set_cors_configuration', pattern='/buckets/{name}/set_cors_configuration'),
     Route(name='bucket_update_versioning', pattern='/buckets/{name}/updateversioning'),
```
kamalgill: yeah, good catch

eucaconsole/views/buckets.py
kamalgill: The corsConfigXML defaults to a sample config when empty, so we'll need the separate hasCorsConfig

eucaconsole/views/buckets.py
kamalgill: Possibly, I'll rework it into a distinct view

eucaconsole/templates/buckets/bucket_details.pt
kamalgill: Doesn't the hasCorsConfig handle that?

eucaconsole/static/js/jasmine-spec/spec_bucket_details.js
```diff
@@ -43,13 +43,32 @@ describe("BucketDetailsPage", function() {
...
+        });
+
+        it("Initial value of deletingCorsConfig is false", function() {
+            expect(scope.deletingCorsConfig).not.toBeTruthy();
+        });
+
     });
 
-    describe("Function initController() Test", function() {
+    describe("Function initController() Tests", function() {
```
kamalgill: I'll make that change.  BTW, could you refer me to the docs or a style guide that mentions this?

eucaconsole/static/js/pages/bucket_details.js
```diff
@@ -82,12 +92,92 @@ angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils'])
                 }
             });
         };
+        $scope.deleteCorsConfig = function ($event) {
+            $event.preventDefault();
+            $scope.deleteError = '';
+            $scope.deletingCorsConfig = true;
+            var deleteDialog = $('#cors-delete-confirmation-modal');
+            var corsForm = $('#cors-deletion-form');
+            var csrfToken = corsForm.find('#csrf_token').val();
```
kamalgill: I'll simplify the jQuery lookup.  Not sure if we could put the token on the scope though, since some pages have multiple csrf_token inputs (as they are auto-generated via WTForms).

eucaconsole/static/js/pages/bucket_details.js
kamalgill: A blog post I read about cross-controller communication recommended this approach, although I can't seem to dig up the page now.  Could you point me to docs or a blog post on how to better handle this?  Should we use $broadcast and $on via $rootScope here?

eucaconsole/views/buckets.py
coyote240: `function hasCorsConfig () {
  corsConfigXml !== corsSampleConfigXml;
}`

eucaconsole/static/js/pages/bucket_details.js
kamalgill: Leveraging $rootScope $broadcast/$on seems to work here

eucaconsole/views/buckets.py
kamalgill: Can't use that approach since it would give a misleading value when the actual CORS config is the same as the sample config.  At any rate, I cleaned up how the boolean is set, moving it client-side.

eucaconsole/static/js/pages/bucket_details.js
kamalgill: Done, thanks for the suggestion!

eucaconsole/static/js/jasmine-spec/spec_bucket_details.js
```diff
@@ -43,13 +43,32 @@ describe("BucketDetailsPage", function() {
         it("Initial value of objectsCountLoading is true", function() {
             expect(scope.objectsCountLoading).toBeTruthy();
         });
+
+        it("Initial value of hasCorsConfig is false", function() {
```
kamalgill: Done

eucaconsole/views/buckets.py
kamalgill: Done

eucaconsole/static/js/jasmine-spec/spec_bucket_details.js
kamalgill: Done

eucaconsole/static/js/pages/bucket_details.js
kamalgill: Done

eucaconsole/views/buckets.py
coyote240: Sounds good, thanks.

dkavanagh:
I noticed DELETE and PUT operations are being used. In tweens.py, the usage_log_tween_factory should be modified to log actions for those verbs in the same way POST is handled.

kamalgill:
@dkavanagh: I added DELETE and PUT operations to the log usage tween as requested.  However, I opted not to add the method to the log statement since that would change the log format and possibly break any parsers users have in place to process the log files, which is too drastic a change for 4.4.  Perhaps we should update the logging tween to include the method (POST, PUT, or DELETE) for the 5.0 release?

eucaconsole/views/buckets.py
```diff
@@ -756,6 +770,14 @@ def bucket_update_versioning(self):
             self.request.error_messages = self.versioning_form.get_errors_list()
         return self.render_dict
 
+    def get_controller_options_json(self):
+        return BaseView.escape_json(json.dumps({
+            'bucket_name': self.bucket.name,
+            'cors_config_xml': self.cors_configuration_xml,
+            'bucket_objects_count_url': self.request.route_path(
+                'bucket_objects_count_versioning_json', name=self.bucket.name),
```
dkavanagh: In the new way, this would not be here. Forgot to flag this last time. Just use a literal on the client side.

PR #2062, opened 2016-09-26
GUI-2758 add logic to keep login buttons disabled unless forms populated

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2758

### DIFF
--- a/eucaconsole/static/js/jasmine-spec/spec_login.js
+++ b/eucaconsole/static/js/jasmine-spec/spec_login.js
@@ -57,29 +57,22 @@ describe("LoginPage", function() {
             expect(scope.eucaNotValid).toBe(true);
         });
 
-        it("should have value of eucaNotValid is false", function() {
+        it("should have value of eucaNotValid is true", function() {
             spyOn(scope, 'eucaLoginNotValid');
-            console.log("account before = "+$("#account").val());
-            $("#account").val("blah");
-            $("#username").val("blah");
-            $("#password").val("blah");
-            console.log("account after = "+$("#account").val());
+            scope.eucaCheckValid();
             timeout.flush();
             expect(scope.eucaLoginNotValid).toHaveBeenCalled();
-            expect(scope.eucaNotValid).toBe(false);
         });
 
         it("should have value of awsNotValid is true", function() {
             expect(scope.awsNotValid).toBe(true);
         });
 
-        it("should have value of awsNotValid is false", function() {
+        it("should have value of awsNotValid is true", function() {
             spyOn(scope, 'awsLoginNotValid');
-            $("#access_key").val("blah");
-            $("#secret_key").val("blah");
+            scope.awsCheckValid();
             timeout.flush();
             expect(scope.awsLoginNotValid).toHaveBeenCalled();
-            expect(scope.awsNotValid).toBe(false);
         });
 
     });
--- a/eucaconsole/static/js/jasmine-spec/spec_login.js
+++ b/eucaconsole/static/js/jasmine-spec/spec_login.js
@@ -58,11 +58,14 @@ describe("LoginPage", function() {
         });
 
         it("should have value of eucaNotValid is false", function() {
+            spyOn(scope, 'eucaLoginNotValid');
+            console.log("account before = "+$("#account").val());
             $("#account").val("blah");
             $("#username").val("blah");
             $("#password").val("blah");
-            console.log("account = "+$("#account").val());
+            console.log("account after = "+$("#account").val());
             timeout.flush();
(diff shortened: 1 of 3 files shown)


eucaconsole/static/js/pages/login.js
```diff
@@ -112,6 +119,38 @@ angular.module('LoginPage', ['EucaConsoleUtils'])
...
+            $("#account").on('keydown', $scope.eucaCheckValid);
+            $("#username").on('keydown', $scope.eucaCheckValid);
+            $("#password").on('keydown', $scope.eucaCheckValid);
+            $("#access_key").on('keydown', $scope.awsCheckValid);
+            $("#secret_key").on('keydown', $scope.awsCheckValid);
+        };
+        $scope.eucaCheckValid = function() {
+            $timeout(function() {  // this causes an additional digest cycle after current thread completes
+                $scope.eucaNotValid = $scope.eucaLoginNotValid();
+            }, 100);
```
coyote240: Give this function a return value and use it in the ng-disabled="eucaCheckValid()" and you can avoid the state variables and $timeout.

eucaconsole/static/js/pages/login.js
```diff
@@ -112,6 +119,38 @@ angular.module('LoginPage', ['EucaConsoleUtils'])
...
+            var password = $('#password').val();
+            if (!account || !username || !password) {
+                return true;
+            }
+            return false;
+        };
+        $scope.awsCheckValid = function() {
+            $timeout(function() {  // this causes an additional digest cycle after current thread completes
+                $scope.awsNotValid = $scope.awsLoginNotValid();
+            }, 100);
```
coyote240: See not for eucaCheckValid above.

eucaconsole/templates/login.pt
```diff
@@ -88,19 +88,19 @@
                             </div>
                             <div class="row">
                                 <div class="small-12 columns">
-                                    <button type="submit" class="submit button" id="euca-login-button" i18n:translate="">
+                                    <a target="_blank" href="${layout.support_url}" i18n:translate="">Forgot your password?</a>
```
coyote240: Known vulnerability with target="_blank".  You can give it a name, such as target="euca-login" or something like that.  Also, add rel="noopener noreferrer" to avoid the exploit:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

eucaconsole/static/js/pages/login.js
```diff
@@ -112,6 +119,38 @@ angular.module('LoginPage', ['EucaConsoleUtils'])
...
+            $("#access_key").on('keydown', $scope.awsCheckValid);
+            $("#secret_key").on('keydown', $scope.awsCheckValid);
+        };
+        $scope.eucaCheckValid = function() {
+            $timeout(function() {  // this causes an additional digest cycle after current thread completes
+                $scope.eucaNotValid = $scope.eucaLoginNotValid();
+            }, 100);
+        };
+        $scope.eucaLoginNotValid = function () {
+            var account = $('#account').val();
```
coyote240: There is a cost to jquery lookups, even on ID.  Could consider memoizing them, or at least initializing the form fields as variables at the top of the controller.

eucaconsole/static/js/jasmine-spec/spec_login.js
```diff
@@ -49,4 +49,29 @@ describe("LoginPage", function() {
...
+
+        it("should have value of eucaNotValid is true", function() {
+            expect(scope.eucaNotValid).toBeTruthy();
+        });
+
+        it("should have value of eucaNotValid is false", function() {
+            $("#account").val("blah");
+            $("#username").val("blah");
+            $("#password").val("blah");
+            expect(scope.eucaNotValid).not.toBeTruthy();
```
coyote240: `expect(scope.eucaNotValid).toBe(true)`, since it's an explicitly boolean value

kamalgill: Shouldn't it be `expect(scope.eucaNotValid).toBe(false)`?

eucaconsole/static/js/jasmine-spec/spec_login.js
```diff
@@ -49,4 +50,32 @@ describe("LoginPage", function() {
...
+
+        it("should have value of eucaNotValid is true", function() {
+            expect(scope.eucaNotValid).toBe(true);
+        });
+
+        it("should have value of eucaNotValid is false", function() {
+            $("#account").val("blah");
+            $("#username").val("blah");
+            $("#password").val("blah");
+            console.log("account = "+$("#account").val());
```
dkavanagh: Note, this logs "account = undefined", which leads me to think the above assignments failed entirely.

PR #2064, opened 2016-09-28
GUI-2288: Switch to new tag editor on security group pages

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2288 by incorporating new tag editor on Create Security Group page and details page

### DIFF
--- a/eucaconsole/views/securitygroups.py
+++ b/eucaconsole/views/securitygroups.py
@@ -254,6 +254,7 @@ class SecurityGroupView(TaggedItemView):
             security_group_names=self.get_security_group_names(),
             controller_options_json=controller_options_json,
             is_vpc_supported=self.is_vpc_supported,
+            tags=self.serialize_tags(self.security_group.tags) if self.security_group else [],
         )
 
     def get_default_vpc_network(self):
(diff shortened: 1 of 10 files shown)


eucaconsole/views/securitygroups.py
```diff
@@ -310,7 +311,8 @@ def securitygroup_create(self):
                     self.add_rules(security_group=new_security_group, traffic_type='egress')
                 if tags_json:
                     tags = json.loads(tags_json)
-                    for tagname, tagvalue in tags.items():
+                    tags_dict = self.normalize_tags(tags)
```
coyote240: Are you sure you need to call normalize_tags?  If I remember right, it was only used in elbs to work around something wicked going on in there.  normalize_tags isn't called in other tag editor implementations.

kamalgill: The tags come across as a list of dicts (after json.loads), so we need the normalize_tags call here.

kamalgill: Note that we also seem to need the normalize_tags call when using the new tag editor on the Create Image from Instance page (see the GUI-2283-new WIP branch)

coyote240: Lame.  Ok, let's merge it.

PR #2071, opened 2016-10-05
OIDC login support

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2666
https://eucalyptus.atlassian.net/browse/GUI-2768


eucaconsole/models/auth.py
```diff
@@ -470,3 +474,72 @@ def authenticate(self, timeout=20):
...
+        self.validate_certs = validate_certs
+        self.kwargs = validate_kwargs
+
+    def authenticate(self, token, account_name, duration=3600, timeout=20):
+        try:
+            creds = self._authenticate_oidc_(token, account_name, self.dns_enabled, duration, timeout)
+        except socket.error as err:
+            # handle case where dns attempt was good, but user unauthorized
+            error_msg = getattr(err, 'msg', None)
+            if error_msg and error_msg == 'Unauthorized' or error_msg == 'Forbidden':
```
coyote240: Is this not an HTTP/S connection?  Can we not check for status codes?

eucaconsole/models/auth.py
```diff
@@ -470,3 +474,72 @@ def authenticate(self, timeout=20):
...
+                raise err
+            elif getattr(err, 'reason', '').find('SSL') > -1:
+                raise err
+            self.dns_enabled = False
+            creds = self._authenticate_oidc_(token, account_name, self.dns_enabled, duration, timeout)
+        return creds
+
+    def _authenticate_oidc_(self, token, account_name, dns_enabled, duration, timeout):
+        if dns_enabled:
+            region = RegionInfo(name='eucalyptus', endpoint='tokens.'+self.host)
```
coyote240: I would either use a string format for the interpolation here, or failing that, put spaces around the + operator.  I'm surprised your linter doesn't complain about this.

eucaconsole/models/auth.py
```diff
@@ -470,3 +474,72 @@ def authenticate(self, timeout=20):
...
+
+        conn = STSConnection(
+            port=self.port,
+            path=auth_path,
+            region=region,
+            https_connection_factory=(conn_factory, ())
+        )
+        conn.http_connection_kwargs['timeout'] = timeout
+        if duration > 3600:  # this call won't allow than 1 hour duration
+            duration = 3600
```
coyote240: duration = min(duration, 3600)

eucaconsole/static/js/pages/login.js
```diff
@@ -40,8 +43,9 @@ angular.module('LoginPage', ['EucaConsoleUtils'])
             if (Modernizr.sessionstorage) {
                 sessionStorage.removeItem('copy-object-buffer');
             }
-            var storedRegion = (Modernizr.localstorage && localStorage.getItem('euca-region')) || '';
+            var storedRegion = (Modernizr.localstorage && localStorage.getItem('euca-region')) || 'euca';
             $("#euca-region").val(storedRegion);
+            $scope.oidcUrl = $scope.oidcLoginLink + "&state=oidc-" + $.base64.encode(storedRegion);
```
coyote240: Can use native btoa() here instead of the jquery base64 encoder, I believe.

eucaconsole/static/js/pages/login.js
```diff
@@ -86,6 +90,8 @@ angular.module('LoginPage', ['EucaConsoleUtils'])
                 var target = $('#aws');
                 target.siblings().removeClass('active').end().addClass('active');
             }
+            var accountName = $.cookie('accountName');
```
coyote240: Maybe consider using the AngularJS $cookies service instead of jquery.

eucaconsole/static/js/pages/login.js
```diff
@@ -157,5 +163,30 @@ angular.module('LoginPage', ['EucaConsoleUtils'])
...
+            $event.preventDefault();
+            $("#oidc-account-modal").foundation('reveal', 'open');
+            $timeout(function() {
+                accountNameField.focus();
+                $scope.oidcCheckValid();
+            }, 250);
+        };
+        $scope.handleOIDCLogin = function($event) {
+            $event.preventDefault();
+            $.cookie('accountName', accountNameField.val(), {expires: 180});
```
coyote240: Same as above, good to checkout AngularJS $cookie service.

eucaconsole/static/js/pages/login.js
```diff
@@ -157,5 +163,30 @@ angular.module('LoginPage', ['EucaConsoleUtils'])
...
+            window.location.href = $scope.oidcUrl + "-" + $('#account-name').val();
```
coyote240: Take a look at AngularJS $location.

eucaconsole/static/sass/pages/login.scss
```diff
@@ -52,6 +52,16 @@
     }
 }
 
+.button.oidc {
+    background-color: $oidc-blue !important;
```
coyote240: Can you increase specificity and avoid !important?  These are a drag to deal with later.

eucaconsole/views/login.py
```diff
@@ -151,6 +203,64 @@ def handle_login(self):
 
         return self.render_dict
 
+    def handle_web_identity_login(self, token):
+        auth = self.get_oidc_authenticator()
+        session = self.request.session
+
+        try:
+            state = role_session_name=token['state']
```
coyote240: For the purpose of self-documentation, it would be cool if these try blocks wrapped a method call instead of a large piece of bare code.

eucaconsole/models/auth.py
dkavanagh: It won't make a difference because the code is 403 in both cases.

eucaconsole/static/sass/pages/login.scss
dkavanagh: maybe I don't need it. I originally did when it was part of another rule.

eucaconsole/models/auth.py
dkavanagh: had to cast duration as int, but yes

eucaconsole/static/js/pages/login.js
coyote240: Oh, yeah, for sure.

eucaconsole/models/auth.py
dkavanagh: I'd have to go back to discussions I had w/ Steve about this. I'll see if I still have those e-mails.

eucaconsole/static/js/pages/login.js
dkavanagh: I found 11 places our code uses the jquery version. I'll open a separate ticket to verify btoa() and remove the jquery dependency. (GUI-2781)

eucaconsole/static/js/pages/login.js
dkavanagh: Agreed. This was just continuing with existing usage.

eucaconsole/static/js/pages/login.js
dkavanagh: The main value of $location is in parsing the URL, but I did try using $location.url() which only allows modification of url parts from the path onward.

eucaconsole/views/login.py
dkavanagh: While I agree that the part of the code throwing the exception is made clear, I'd also argue that inserting a bunch of exception handling code in the middle of what was a linear code sequence makes that a little harder to follow. I went with the philosophy that exceptions are exceptional and keeping the main flow together was more intuitive to read. Perhaps a comment in the code instead? I'll hold on to the change locally till we talk.

coyote240: I generally look for code flow to have a narrative, and functions aid in that narrative, like:
```python
try:
    code_that_may_fail()
catch Exception as e:
    handle_the_failure(e)
    #or other linear handling
```

Meaning, when I encounter the overall flow of this portion of the program, I'm looking for functions named such that I know what is happening, rather than a large block of implementation code that tells me how it is happening.

eucaconsole/static/sass/includes/_eucavariables.scss
```diff
@@ -50,6 +50,7 @@ $hpe-bronze: #80746E;
 $hpe-steel: #5F7A76;
 $hpe-orange: #FF8D6D;
 
+$oidc-blue: #3560A0;
```
kamalgill: Is there a compelling reason to prefix this variable with 'oidc-'?

dkavanagh: because it's a color that is specifically for this feature, not from the themed colors (per Jenny)

dkavanagh: I could see keeping this in the login.scss file, but put it here because we've put color defs here.

eucaconsole/views/login.py
kamalgill: I'm with Adam here.  Let's refactor the 'try' block to perhaps call two methods... one for authenticating and another for setting the session, and only wrap the authenticate call in the try block?

dkavanagh: Should we do the same for all 3 methods for login? Same situation with the other 2 existing ones.

eucaconsole/views/login.py
```diff
@@ -121,6 +134,8 @@ def __init__(self, request):
             login_refresh=self.login_refresh,
             came_from=self.came_from,
             controller_options_json=options_json,
+            oidc_enabled=self.oidc_host is not None,
+            oidc_link_text=self.request.registry.settings.get('oidc.login.button.label', 'oidc login')
```
kamalgill: Should we mark the fallback button label ('oidc login') for i18n?

eucaconsole/templates/login.pt
```diff
@@ -179,6 +186,30 @@
                 <a class="close-reveal-modal">&#215;</a>
             </div>
         </div>
+        <div>
```
kamalgill: Minor point, but we don't need the empty wrapper div here

eucaconsole/views/roles.py
```diff
@@ -216,6 +216,11 @@ def _get_trusted_entity_(parsed_policy):
             if isinstance(svc, list):
                 svc = svc[0]
             return _(u'Service ') + svc
+        elif 'Federated' in principal.keys():
+            fed = principal['Federated']
+            if isinstance(fed, list):
+                fed = fed[0]
+            return _(u'Federated ') + fed
```
kamalgill: Do we want to mark this for i18n?  Is it displayed to the user, or is it for internal use only when returning the trusted entity?

dkavanagh: It is translated because it is displayed to the user.

PR #2085, opened 2016-10-08
GUI-2778: Fix bucket choices in Create Image from Instance page for Angular 1.5 compatibility

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2778

### DIFF
--- a/eucaconsole/templates/instances/instance_create_image.pt
+++ b/eucaconsole/templates/instances/instance_create_image.pt
@@ -37,7 +37,7 @@
                         <p i18n:translate="">
                             Specify where your image should be stored in the cloud.
                         </p>
-                        ${panel('form_field', field=create_image_form['s3_bucket'], leftcol_width=3, rightcol_width=9, ng_attrs={'model': 'bucketName', 'options': 'k as v for (k, v) in bucketNameChoices'})}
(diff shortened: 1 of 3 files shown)


eucaconsole/templates/instances/instance_create_image.pt
```diff
@@ -37,7 +37,7 @@
                         <p i18n:translate="">
                             Specify where your image should be stored in the cloud.
                         </p>
-                        ${panel('form_field', field=create_image_form['s3_bucket'], leftcol_width=3, rightcol_width=9, ng_attrs={'model': 'bucketName', 'options': 'k as v for (k, v) in bucketNameChoices'})}
+                        ${panel('form_field', field=create_image_form['s3_bucket'], leftcol_width=3, rightcol_width=9, ng_attrs={'model': 'bucketName', 'options': 'k as v for (k, v) in bucketNameChoices track by v'})}
```
dkavanagh: Does this work OK? We've tried to avoid select as with track by per Angular's documentation.

kamalgill: Yes, this seems to work here, and re-working it to use item id/label approach isn't feasible due to the inline create bucket dialog, which is reused elsewhere.

PR #2086, opened 2016-10-11
DOC-1988: update to include enabling CORS.

tvo318:
@dkavanagh: changes reviewed and approved. Please merge to 4.4 codebase. Thanks!

### DIFF
--- a/eucaconsole/static/html/help/console_bucket_detail.html
+++ b/eucaconsole/static/html/help/console_bucket_detail.html
@@ -102,7 +102,7 @@
 
             <ol class="ol steps"><li class="li step stepexpand">
                     <span class="ph cmd">To edit the sharing settings associated with bucket Access Control Lists (ACLs):</span>
-                    <div class="itemgroup info"><div class="note note"><span class="notetitle">Note:</span> For more information about ACLs, see <a class="xref" href="../admin-guide/manage_resources_walrus.html" title="This topic explains Walrus resources.">Manage Walrus Resources</a>.</div>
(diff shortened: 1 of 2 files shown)


eucaconsole/static/html/help/console_bucket_detail.html
```diff
@@ -86,10 +98,13 @@ <h2 class="title topictitle2">Edit the sharing settings</h2>
...
         <div class="body taskbody">
             <div class="section context">The sharing settings dictate whether your account or another account can
-                acccess your bucket and its contents.</div>
+                access your bucket and its contents.</div>
 
-            <ol class="ol steps"><li class="li step">
-                    <span class="ph cmd">To edit the sharing settings:</span>
+            <ol class="ol steps"><li class="li step stepexpand">
+                    <span class="ph cmd">To edit the sharing settings associated with bucket Access Control Lists (ACLs):</span>
+                    <div class="itemgroup info"><div class="note note"><span class="notetitle">Note:</span> For more information about ACLs, see <a class="xref" href="../admin-guide/manage_resources_walrus.html" title="This topic explains Walrus resources.">Manage Walrus Resources</a>.</div>
```
tvo318: You are right, David.. The one on this Help page doesn't work either. I will use an absolute path. We will have to update it for every release, unfortunately, because the release version appears in URL.

dkavanagh: Do the docs support the use of "latest" in place of version # in the url? That's commonly used in other places and it would solve the problem here.

tvo318:
@dkavanagh, files are now corrected with absolute paths to link to other docs outside of the Help.

dkavanagh:
Thanks. Good catch on the instance detail link.

PR #2089, opened 2016-10-13
Gui 2643 moto test improvements

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2643

There is a repeated function "setup_session()" which would be nice to have in a base-class. I'll look at doing that next.


tests/elbs/test_elbs.py
```diff
@@ -109,42 +108,54 @@ def test_elb_monitoring_tab_view(self):
...
-        health_check = Mock(
-            target='HTTPS:443/', interval=300, timeout=15, healthy_threshold=3, unhealthy_threshold=5)
-        elb.health_check = health_check
-        request = self.create_request()
-        view = ELBHealthChecksView(request, elb=elb).elb_healthchecks()
-        form = view.get('elb_form')
-        self.assertEqual(form.ping_protocol.data, 'HTTPS')
-        self.assertEqual(form.ping_port.data, 443)
-        self.assertEqual(form.ping_path.data, '/')
+# commented out because moto doesn't support DescribeLoadBalancerPolicies, which the BaseELBView needs.
```
coyote240: Can we use @unittest.skip instead?
https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

tests/volumes/test_volumes.py
```diff
@@ -196,27 +195,30 @@ class MockVolumeViewTestCase(BaseViewTestCase, MockVolumeMixin):
...
         self.assertEqual(view.get('volume_name'), volume.id)
 
-    @mock_ec2
-    def test_volume_detail_view_with_new_volume(self):
-        volume, conn = self.make_volume()
-        request = self.create_request(matchdict=dict(id='new'))
-        view = VolumeView(request, ec2_conn=conn).volume_view()
-        self.assertEqual(view.get('volume'), None)
+# commented out because moto connection does not have "host" attribute that this VolumeView requires
+#    @mock_ec2
```
coyote240: Same as above, @unittest.skip

tests/buckets/test_buckets.py
```diff
@@ -121,54 +120,64 @@ def test_upload_post_returns_400_when_file_uploads_config_is_disabled(self):
...
 
     @mock_s3
+    @mock_ec2
     def test_bucket_details_view_without_versioning(self):
         request = self.create_request()
-        bucket, bucket_acl = self.make_bucket()
-        view = BucketDetailsView(request, bucket=bucket, bucket_acl=bucket_acl).bucket_details()
+        self.setup_session(request)
+        request.matchdict['name'] = 'test_bucket'
+        self.make_bucket()
```
kamalgill: This line doesn't appear to do anything, so probably safe to remove.

tests/alarms/test_alarms.py
```diff
@@ -66,14 +65,16 @@ class ResourceAlarmsTestCase(unittest.TestCase, MockAlarmMixin):
     @mock_cloudwatch
     def test_fetch_alarms_for_instance(self):
         instance_id = 'i-123456'
-        cw_conn, alarm_created = self.create_alarm(dimensions={'InstanceId': [instance_id]})
+        alarm_created = self.create_alarm(dimensions={'InstanceId': [instance_id]})
```
kamalgill: PyCharm flags alarm_created as an unused variable, so the line here (and elsewhere) could be simplified to `self.create_alarm(dimensions={'InstanceId': [instance_id]})`

tests/instances/test_instances.py
```diff
@@ -319,29 +318,37 @@ def test_deprecated_required_validator(self):
...
 class InstanceMonitoringViewTestCase(BaseViewTestCase, MockInstanceMixin):
 
+    @mock_ec2
     def test_instance_monitoring_view_duration_choices(self):
         request = testing.DummyRequest()
+        self.setup_session(request)
         instance = self.make_instance()
-        view = InstanceMonitoringView(request, instance=instance).instance_monitoring()
+        request.matchdict['id'] = instance.id
+        request.is_xhr = True;
```
kamalgill: No need for trailing semicolon here

tests/buckets/test_buckets.py
dkavanagh: It totally does something. It creates a bucket.

PR #2098, opened 2016-10-19
GUI-2794 initial reporting page w/ tabs

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2794

### DIFF
--- a/eucaconsole/static/js/jasmine-spec/spec_reporting.js
+++ b/eucaconsole/static/js/jasmine-spec/spec_reporting.js
@@ -8,25 +8,53 @@
 
 describe("Reporting", function() {
 
-    beforeEach(angular.mock.module('Reporting'));
+    beforeEach(angular.mock.module('ReportingPage'));
 
-    var $scope, $location;
-    beforeEach(angular.mock.inject(function(_$location_, _$rootScaope_) {
+    var $rootScope, $location, $compile;
+    beforeEach(angular.mock.inject(function(_$location_, _$rootScope_, _$compile_) {
+        $rootScope = _$rootScope_;
         $location = _$location_;
-        $scope = _$rootScope_.$new();
+        $compile = _$compile_;
     }));
 
+    var controller;
     beforeEach(function() {
-        var template = window.__html__['templates/reporting/reporting.pt'];
-        setFixtures(template);
+        var element = $compile(
+            '<dl navigation="" reporting-configured="true"><dd></dd><dd></dd></dl>'
+        )($rootScope);
+        $rootScope.$digest();
+
+        $location.path('/reporting/dashboard');
+        $rootScope.$apply();
+
+        controller = element.controller('navigation');
     });
 
     describe("initial tab test", function() {
         it("should set preferences path", function() {
-            $location.path('/reporting/dashboard');
-            expect(location.path() == '/reporting/preferences').toBe(false);
-            scope.setInitialTab('false');
-            expect(location.path() == '/reporting/preferences').toBe(true);
+            expect($location.path() == '/reporting/preferences').toBe(false);
+            controller.setInitialTab('false');
+            expect($location.path() == '/reporting/preferences').toBe(true);
+        });
+    });
+
+    describe("active tab test", function() {
+        it("should set active tab state", function() {
+            expect(controller.isTabActive('/dashboard') == 'active').toBe(true);
+            expect(controller.isTabActive('/reports') == 'active').toBe(false);
+            expect(controller.isTabActive('/preferences') == 'active').toBe(false);
+            $location.path('/reporting/reports');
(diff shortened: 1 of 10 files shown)


eucaconsole/static/js/jasmine-spec/spec_reporting.js
```diff
@@ -0,0 +1,60 @@
...
+        it("should set preferences path", function() {
+            expect($location.path() == '/reporting/preferences').toBe(false);
+            controller.setInitialTab('false');
+            expect($location.path() == '/reporting/preferences').toBe(true);
+        });
+    });
+
+    describe("active tab test", function() {
+        it("should set active tab state", function() {
+            expect(controller.isTabActive('/dashboard') == 'active').toBe(true);
```
coyote240: This would be good to break up.  It's not important that every case is checked.  I'd choose one set and be done.

Tests can be concise, and it's okay to have a few.

eucaconsole/static/js/jasmine-spec/spec_reporting.js
```diff
@@ -0,0 +1,60 @@
...
+            expect(controller.isTabActive('/preferences') == 'active').toBe(false);
+            $location.path('/reporting/reports');
+            expect(controller.isTabActive('/dashboard') == 'active').toBe(false);
+            expect(controller.isTabActive('/reports') == 'active').toBe(true);
+            expect(controller.isTabActive('/preferences') == 'active').toBe(false);
+            $location.path('/reporting/preferences');
+            expect(controller.isTabActive('/dashboard') == 'active').toBe(false);
+            expect(controller.isTabActive('/reports') == 'active').toBe(false);
+            expect(controller.isTabActive('/preferences') == 'active').toBe(true);
+            $location.path('/reporting/blah');
```
coyote240: This should be a separate test, for example ```it('should go to default tab when incorrect path is specified')``` or something like that.

eucaconsole/static/js/jasmine-spec/spec_reporting.js
```diff
@@ -0,0 +1,60 @@
...
+        )($rootScope);
+        $rootScope.$digest();
+
+        $location.path('/reporting/dashboard');
+        $rootScope.$apply();
+
+        controller = element.controller('navigation');
+    });
+
+    describe("initial tab test", function() {
```
coyote240: current naming would have this as #setInitialTab.

eucaconsole/static/js/jasmine-spec/spec_reporting.js
```diff
@@ -0,0 +1,60 @@
...
+        it("should set preferences path", function() {
+            expect($location.path() == '/reporting/preferences').toBe(false);
+            controller.setInitialTab('false');
+            expect($location.path() == '/reporting/preferences').toBe(true);
+        });
+    });
+
+    describe("active tab test", function() {
```
coyote240: Current naming would have this as #isTabActive.

eucaconsole/static/js/pages/reporting.js
```diff
@@ -0,0 +1,69 @@
...
+                if (path.indexOf(name) > -1)
+                    return 'active';
+                // this handles case where url path contains something other than reports or prefs, but
+                // falls back to dashboard tab due to router .otherwise clause
+                if (name === '/dashboard' &&
+                    ['/reports', '/preferences'].every(function(val) { return path.indexOf(val) === -1; }) )
+                    return 'active';
+                return '';
+            };
+            this.setInitialTab = function(reportingConfigured) {
```
coyote240: Does this method name describe what is being done?  I would expect it to allow me to set any tab as initial.  Is it more accurate to say that it is forcing preferences, or something?

eucaconsole/templates/reporting/reporting.pt
```diff
@@ -0,0 +1,42 @@
...
+            <metal:crumbs metal:fill-slot="crumbs">
+                <li class="current"><a href="#" i18n:translate="">Usage Reporting</a></li>
+            </metal:crumbs>
+        </metal:breadcrumbs>
+        <!-- Notifications -->
+        <metal:block metal:use-macro="layout.global_macros['notifications']" />
+        <div class="large-12 columns">
+            <dl navigation="" reporting-configured="${reporting_configured}" class="tabs">
+                <dd ng-class="nav.isTabActive('/dashboard')">
+                    <a href="${request.route_path('reporting', subpath='/dashboard')}" i18n:translate="">Dashboard</a>
```
coyote240: A relative path would be fine here, no need for request.route_path.

eucaconsole/static/js/pages/reporting.js
dkavanagh: I think the method name is accurate and param name doesn't lead you to think you pass a tab name.

PR #2101, opened 2016-10-24
GUI-2738 bucket tagging in details page

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2738

### DIFF
--- a/eucaconsole/static/js/pages/bucket_details.js
+++ b/eucaconsole/static/js/pages/bucket_details.js
@@ -7,7 +7,7 @@
  */
 
 /* Bucket details page includes the S3 Sharing Panel */
-angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils', 'CorsServiceModule', 'ModalModule'])
+angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils', 'CorsServiceModule', 'TagEditorModule', 'ModalModule'])
     .controller('BucketDetailsPageCtrl', function ($scope, $rootScope, $http, eucaHandleErrorS3,
                                                    eucaUnescapeJson, CorsService) {
         $http.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
@@ -44,6 +44,16 @@ angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils', 'Cors
                 eucaHandleErrorS3(oData, status);
             });
         };
+        // True if there exists an unsaved key or value in the tag editor field
+        $scope.existsUnsavedTag = function () {
+            var hasUnsavedTag = false;
+            $('input.taginput[type!="checkbox"]').each(function(){
+                if ($(this).val() !== '') {
+                    hasUnsavedTag = true;
+                }
+            });
+            return hasUnsavedTag;
+        };
         $scope.revealModal = function (action) {
             var modal = $('#' + action + '-modal');
             modal.foundation('reveal', 'open');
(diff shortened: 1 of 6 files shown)


eucaconsole/static/js/widgets/tag-editor/tag-editor.js
```diff
@@ -75,7 +75,7 @@ angular.module('TagEditorModule', ['EucaConsoleUtils'])
             }],
             link: function (scope, element, attrs, ctrl, transcludeContents) {
                 var content = transcludeContents();
-                var tags = JSON.parse(content.text() || '{}');
+                var tags = JSON.parse(content.text() || '[]');
```
dkavanagh: This was a bug since tags.filter (in the line below) will fail.

eucaconsole/views/images.py
```diff
@@ -430,7 +430,7 @@ def __init__(self, request):
         if self.image is not None:
             tags = self.serialize_tags(self.image.tags)
         else:
-            tags = '{}'
+            tags = '[]'
```
dkavanagh: This was also a bug since the tag editor needs an array.

eucaconsole/views/buckets.py
```diff
@@ -790,6 +798,43 @@ def get_logging_status(self):
...
+                    self.bucket.set_tags(alltags)
+                except S3ResponseError as err:
+                    if err.status != 200:
+                        raise err
+
+    def remove_tags(self):
+        if self.bucket:
+            self.bucket.delete_tags()
+
+    def serialize_tags(self, tags):
```
kamalgill: This should be a static method, since it overrides a static method in TaggedItemView.  Also, we can use itertools.chain to flatten the list of lists, simplifying this method to the following...

```
    @staticmethod
    def serialize_tags(tags):
        serialized_tags = []
        if tags is not None:
            tags = chain.from_iterable(tags)
            for tag in tags:
                serialized_tags.append({
                    'name': tag.key,
                    'value': tag.value
                })
        return BaseView.escape_json(json.dumps(serialized_tags))
```

eucaconsole/templates/buckets/bucket_details.pt
```diff
@@ -100,6 +101,9 @@
                     <hr />
                     ${panel('s3_sharing_panel', bucket_object=bucket, sharing_form=sharing_form, cors_configuration_xml=cors_configuration_xml)}
                     <hr />
+                    <div ng-cloak="">
+                        <tag-editor ng-model="tags" tal:content="tags | structure"></tag-editor>
+                    </div>
```
kamalgill: Need an hr tag here to separate the tag editor from the submit button, similar to how other detail pages include the tag editor (and per wireframe spec).

eucaconsole/static/js/pages/bucket_details.js
```diff
@@ -44,6 +44,16 @@ angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils', 'Cors
                 eucaHandleErrorS3(oData, status);
             });
         };
+        // True if there exists an unsaved key or value in the tag editor field
+        $scope.existsUnsavedTag = function () {
```
coyote240: Would it be easier to check the values on the $scope than traverse the form?

dkavanagh: perhaps. I just did copy/paste of this from other screens doing the same thing. Maybe open another ticket to fix everywhere?

coyote240: Sounds good.

PR #2103, opened 2016-10-24
Allow override of 1 hour admin session timeout

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2740


kamalgill:
Looks good to me

PR #2105, opened 2016-10-25
GUI-2767, add rel to anchors that open to target _blank

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-2767


dkavanagh:
Weren't we also supposed to give "target" a more meaningful name?

coyote240:
From a security standpoint, turns out it is not important.  We can give it a more relevant name, though it can be difficult to understand from the surrounding code what that should be.

dkavanagh:
As long as it isn't important from a security standpoint.

PR #2108, opened 2016-10-26
GUI-2782 fixes for metrics charts

dkavanagh:
NOTE: Contains PR#2107 as well, merge only after that PR has been merged.

https://eucalyptus.atlassian.net/browse/GUI-2782


eucaconsole/views/cloudwatchapi.py
```diff
@@ -288,7 +288,7 @@ def get_stats_series(self, dimensions=None, label=None):
         multiplier = 1
         divider = 1
         unit = self.unit
-        period = int(self.request.params.get('period', 300))
+        period = int(self.request.params.get('period', '300'))
```
kamalgill: @dkavanagh No need to fall back to '300' if it's converted to an int anyway

dkavanagh: sure, that could be. I saw another error with this line that was a result of the parameter being wrong, but thought the int conversion on an int didn't feel right.

PR #2109, opened 2016-10-26
GUI-2770 modified font paths, removed Oswald and Lato fonts

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2770


dkavanagh:
I'll have to check. I moved the files, found out they weren't referenced anywhere, then removed them, all using git commands, so something didn't end up right.

PR #2113, opened 2016-10-27
GUI-2811, Certificate Editor, select existing cert

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-2811


eucaconsole/static/js/widgets/listener-editor/certificate-editor.js
```diff
@@ -0,0 +1,109 @@
+angular.module('ELBCertificateEditorModule', ['ModalModule'])
+.directive('certificateEditor', function () {
+    return {
+        restrict: 'E',
+        require: {
+            stepData: '?^^stepData'
```
dkavanagh: woah, what's that notation?

coyote240: Check it out in the docs for $compile.  This is requiring that the stepData directive is present on a parent element, but do not fail if it is not, returning null instead.

PR #2116, opened 2016-11-03
GUI-2739: Instance Termination Protection

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2739

### DIFF
--- a/eucaconsole/templates/instances/instance_view.pt
+++ b/eucaconsole/templates/instances/instance_view.pt
@@ -103,33 +103,33 @@
                     <!--! Modal dialogs are near the bottom of this template -->
                     <h6 i18n:translate="">Instance</h6>
                     <div class="row controls-wrapper readonly" ng-cloak="">
-                        <div class="large-2 small-4 columns"><label i18n:translate="" id="instance-status-label">Status</label></div>
-                        <div class="large-10 small-8 columns value" ng-cloak="">
+                        <div class="large-3 small-4 columns"><label i18n:translate="" id="instance-status-label">Status</label></div>
+                        <div class="large-9 small-8 columns value" ng-cloak="">
                             <span class="label radius status {{ instanceState }}">{{instanceState}}</span>
                             <span class="busy" ng-show="isUpdating"></span>
                         </div>
                     </div>
                     <div>
-                        ${panel('form_field', field=instance_form.name, pattern=layout.tag_pattern_value)}
+                        ${panel('form_field', field=instance_form.name, pattern=layout.tag_pattern_value, leftcol_width_large=3, rightcol_width_large=9)}
                     </div>
                     <div tal:condition="instance.root_device_type != 'ebs'" tal:omit-tag="">
                         <div class="row controls-wrapper readonly">
-                            <div class="large-2 small-4 columns"><label i18n:translate="">Instance type</label></div>
-                            <div class="large-10 small-8 columns value">${instance.instance_type}</div>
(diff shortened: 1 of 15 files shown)


eucaconsole/templates/instances/instance_view.pt
```diff
@@ -346,6 +367,53 @@
...
+                Disabling instance termination protection will allow termination of the instance via the console,
+                command-line tools, and client libraries.
+            </p>
+            <p i18n:translate="" tal:condition="not: termination_protection_on">
+                Enabling instance termination protection will prevent termination of the instance via the console,
+                command-line tools, and client libraries.
+            </p>
+            <p i18n:translate="" tal:condition="not: termination_protection_on and scaling_group">
+                NOTE: An instance with termination protection enabled in a scaling group will not prevent the instance
+                from being terminated by a scaling policy.
```
dkavanagh: fwiw, an instances can also terminate itself with halt or shutdown.

eucaconsole/views/instances.py
```diff
@@ -165,6 +165,20 @@ def get_monitoring_state(self, instance=None):
...
+        Get termination protection state as a tuple of state (bool) and label
+        :param instance: Boto EC2 instance object
+        :return: tuple of is_enabled (True or False) and label ('Enabled' or 'Disabled')
+        :rtype: tuple
+        """
+        if not instance:
+            return False
+        termination_protection_attr = self.conn.get_instance_attribute(instance.id, 'disableApiTermination')
+        is_enabled = termination_protection_attr.get('disableApiTermination', False)
+        state_label = _('Enabled') if is_enabled else _('Disabled')
```
dkavanagh: The label is only used in 1 place below. I feel like this if/else can be right there instead of adding the complexity of returning multiple values from what could be a 1 or 2 line function (like in the instances form code).

eucaconsole/views/instances.py
```diff
@@ -165,6 +165,20 @@ def get_monitoring_state(self, instance=None):
...
             return _(u'Detailed') if instance.monitoring_state == 'enabled' else _(u'Basic')
 
+    def get_termination_protection_state(self, instance=None):
+        """
+        Get termination protection state as a tuple of state (bool) and label
+        :param instance: Boto EC2 instance object
+        :return: tuple of is_enabled (True or False) and label ('Enabled' or 'Disabled')
+        :rtype: tuple
+        """
+        if not instance:
```
dkavanagh: Shouldn't the instance view fail fast if the instance can't be found? Throwing a 404 instead of None checks is cleaner.

eucaconsole/templates/instances/instance_view.pt
kamalgill: Not sure what you mean here -- this notice is limited to an instance in a scaling group.

eucaconsole/views/instances.py
kamalgill: Yeah, good point.  Probably safer to remove the instance check altogether here, as a None or 404 check shouldn't be in this method anyway.

eucaconsole/templates/instances/instance_view.pt
kamalgill: Ah, I see the problem here.  The scaling group check is ignored when the conditions are combined and termination protection is on -- will fix.

kamalgill: I added "...or a health check." to the sentence to communicate that an instance with termination protection enabled can still be terminated by an instance or ELB health check failure in a scaling group.

PR #2121, opened 2016-11-11
GUI-2809 add interactions between zone/subnet selection and instance selection

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2809


eucaconsole/templates/elbs/wizard/instances.pt
```diff
@@ -33,13 +33,14 @@
...
-                    ng-model="vpcSubnets" focus-on-load="true" required=""
-                    ng-options="subnet as subnet.label for subnet in vpcSubnetChoices track by subnet.id"
+                    ng-model="instances.vpcSubnets" focus-on-load="true" required=""
+                    ng-options="subnet as subnet.label for subnet in instances.vpcSubnetChoices track by subnet.id"
                     data-placeholder="Select VPC subnets" i18n:attributes="data-placeholder">
                 </select>
             </div>
             <span id="hidden_vpc_subnet_empty_option" class="hide" i18n:translate="">No subnets found</span>
         </div>
+        <div>&nbsp;</div>
```
coyote240: What is this for?

eucaconsole/static/js/pages/elb-wizard/wizard.js
```diff
@@ -96,6 +105,57 @@ angular.module('ELBWizard', [
         }
     };
 })
+.directive('fetchData', function(InstancesService, ZonesService, VPCService, ELBWizardService, eucaHandleError) {
```
coyote240: This feels weird as an element.  The more I look at it I think it can be put into the main controller.

eucaconsole/static/js/pages/elb-wizard/wizard.js
```diff
@@ -96,6 +105,57 @@ angular.module('ELBWizard', [
...
+    return {
+        restrict: 'E',
+        link: function(scope, elem, attrs) {
```
coyote240: If we do keep this directive, there is no reason to avoid an isolate scope.  You just don't expect data lookups in a link method, and it seems the only reason we're doing this in the link is to avoid an isolate scope. 

Again, if this code does not move to the main controller, I think this code should move to the directive controller with an isolate scope.

eucaconsole/static/js/pages/elb-wizard/instances.js
```diff
@@ -7,24 +7,91 @@
...
+        vm.instancesLoading = (ELBWizardService.values.instances === undefined);
+        // using $watch instead of ng-change because...
+        // https://github.com/leocaseiro/angular-chosen/issues/145
+        $scope.$watch('instances.availabilityZones', function(newval, oldval) {
+            if (newval === oldval) return;  // leave unless there's a change
+            vm.handleDeselectionDueToZones(newval, oldval);
+        });
+        vm.handleDeselectionDueToZones = function(newval, oldval) {
+            var zoneDiff = oldval.filter(function(x) {
+                var idx = newval.findIndex(function(val) { return val.id === x.id; });
```
dkavanagh: I was thinking about that earlier, then forgot to do it.

eucaconsole/static/js/pages/elb-wizard/instances.js
```diff
@@ -7,24 +7,91 @@
...
+                if (instanceInSubnet) instance.selected = false;
+            });
+        };
+        // may want to make this event-driven by having instance selector use callback upon selection change
+        $scope.$watch('instances.instances', function(newval, oldval) {
+            if (newval === oldval) return;  // leave unless there's a change
+            vm.handleInstanceSelectionChange(newval, oldval);
+        }, true);
+        vm.handleInstanceSelectionChange = function(newval, oldval) {
+            if (vm.vpcNetwork === 'None') {
```
dkavanagh: I was thinking about this earlier, but forgot to actually do it.

eucaconsole/templates/elbs/wizard/instances.pt
dkavanagh: oh, cruft from old DOM. Could update CSS instead.

eucaconsole/static/js/pages/elb-wizard/instances.js
dkavanagh: When I really consider this, I end up with a function that has 2 other functions passed, to implement the code on 70/84 and 72/86. Not sure it becomes any easier to reason about with that much abstraction. Yes, less duplicated code, but if it were just passing different variables I'd totally agree.

PR #2122, opened 2016-11-14
GUI-2818: Implement Network tab for new ELB wizard

kamalgill:
https://eucalyptus.atlassian.net/browse/GUI-2818


eucaconsole/static/js/pages/elb-wizard/wizard.js
```diff
@@ -131,6 +132,15 @@ angular.module('ELBWizard', [
...
                     });
+                VPCService.getVPCSecurityGroups().then(
+                    function success(result) {
+                        result.forEach(function(val) {
+                            ELBWizardService.values.vpcSecurityGroupChoices.push(val);
+                        });
+                    },
+                    function error(errData) {
+                        eucaHandleError(errData.data.message, errData.status);
+                    });
```
kamalgill: @dkavanagh: Note that I was leveraging the vpcsecuritygroups_json route you added to routes.py (in commit https://github.com/eucalyptus/eucaconsole/commit/620c3a025c72b9a) , as well as the getVPCSecurityGroups method you added to vpc-service.js (see https://github.com/eucalyptus/eucaconsole/commit/620c3a025c72b9a837aae1f71e21701234856d9d#diff-2495b6e898fac8221964380d3f385ca0R34).  I'm assuming the distinct route was added since the shape of the response is distinct from security group fetches elsewhere.

eucaconsole/static/js/pages/elb-wizard/network.js
```diff
@@ -1,4 +1,29 @@
...
+    vm.vpcSecurityGroupChoices = ELBWizardService.values.vpcSecurityGroupChoices;
+
+    // Set initial VPC network choice
+    if (vm.vpcNetworkChoices.length) {
+        vm.vpcNetwork = vm.vpcNetworkChoices[0];
+    }
+
+    // Set initial security groups choice to 'default' group
+    if (vm.vpcSecurityGroupChoices.length) {
+        vm.securityGroups = vm.vpcSecurityGroupChoices.filter(function (sgroup) {
```
dkavanagh: I commented in ticket, but should also comment in context. replacing the array vs pushing causes bindings to the original array to be lost. So, summary pane won't be able to display updates to security groups because it binds to the shared model array.
Further, doing this here causes a race condition. If the user clicks to the 2nd tab quickly enough, this controller is initialized before the group choices are loaded, which results in no default setting. Acting on the data we lazily load should happen where the data is loaded to ensure the data is available.

eucaconsole/static/js/pages/elb-wizard/wizard.js
dkavanagh: Agreed, due to the additional API calls made to fetch rules in the securitygroups_json route.

PR #2124, opened 2016-11-16
Gui 2814 added create button.

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2814


eucaconsole/static/js/services/elb-service.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * Copyright 2016 Hewlett Packard Enterprise Development LP
+ *
+ * @fileOverview factory method for elb XHR calls
+ * @requires AngularJS
+ *
+ */
+angular.module('ELBServiceModule', [])
```
coyote240: Is this service to be used outside of the ELB Wizard?  If not, it can be part of the ELBWizard module.

eucaconsole/static/js/services/elb-service.js
```diff
@@ -0,0 +1,46 @@
...
+.factory('ELBService', ['$http', function ($http) {
+    $http.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
```
coyote240: It would be nice to see this sent as a posted json body instead of an encoded URL.

eucaconsole/templates/elbs/wizard/advanced.pt
```diff
@@ -1,4 +1,4 @@
-<form>
+<form id="elb-form">
```
coyote240: Why ids?

eucaconsole/static/sass/pages/elb_wizard.scss
```diff
@@ -77,3 +77,7 @@
 #controls_vpc_subnet {
     padding-bottom: 1.25rem;
 }
+
```
coyote240: Styling against ids is troublesome.

eucaconsole/static/js/pages/elb-wizard/advanced.js
```diff
@@ -1,7 +1,16 @@
 angular.module('ELBWizard')
-.controller('AdvancedController', ['$scope', '$routeParams', function ($scope, $routeParams) {
-    this.protocol = 'HTTP';
-    this.port = 80;
-    this.path = '/';
+.controller('AdvancedController', ['$scope', '$routeParams', 'ELBWizardService', 'ELBService', function ($scope, $routeParams, ELBWizardService, ELBService) {
+    this.values = ELBWizardService.values;
```
coyote240: This tends to make the code opaque, and also exposes all the values of the ELB Wizard to this view, which is not necessary.  Why not enumerate the values to be visible to this tab instead?  I would like that better for sure.

eucaconsole/static/js/services/elb-service.js
dkavanagh: I anticipated other service calls, not just this.

eucaconsole/templates/elbs/wizard/advanced.pt
dkavanagh: This was to leverage css that existed for this tag.

eucaconsole/static/js/pages/elb-wizard/advanced.js
dkavanagh: In this case, I did this to allow for values that weren't arrays. Binding to values lets me see changes of simple values within it, like pingPort.

eucaconsole/static/sass/pages/elb_wizard.scss
dkavanagh: If this is for spacing in that particular scenario, how is it troublesome?

aliceh:
Retest this please

eucaconsole/static/sass/pages/elb_wizard.scss
coyote240: Let's let this one ride for now.  I do think we would benefit from revisiting our semantic markup strategy in the future.

PR #2126, opened 2016-11-18
GUI-2831 - ELB Wizard directive

coyote240:
https://eucalyptus.atlassian.net/browse/GUI-2831


eucaconsole/static/js/pages/elb-wizard/wizard.js
```diff
@@ -101,6 +117,9 @@ angular.module('ELBWizard', [
         },
 
         displaySummary: function(step) {
+            if(!this.nav) {
+                return;
```
dkavanagh: Thinking this should return explicit truth value. I guess undefined == false, but I like explicitness.

eucaconsole/templates/elbs/wizard/main.pt
```diff
@@ -32,9 +32,9 @@
         </h3>
 
         <div class="large-8 columns">
-            <div class="wizard">
+            <div class="wizard" ng-app="ELBWizard" elb-wizard="" vpc-enabled="${ is_vpc_supported }" cloud-type="${ cloud_type }">
```
dkavanagh: ng-app is defined below as well. Is that by accident?

eucaconsole/static/js/pages/elb-wizard/wizard.js
coyote240: I tend to agree in Python, less so in JS, in general.  Note this next time you write a function that returns like ```return this.value && this.oldValue``` which will return this.oldValue.  That being said, I don't mind adding ```false.```

coyote240: Oh.  I can use line 123 of this very file as an example.

PR #2129, opened 2016-12-02
GUI-2844: Move non-configurable and developer-only settings out of default console.ini

kamalgill:
Also, improve comments for various console.ini properties.

Implements https://eucalyptus.atlassian.net/browse/GUI-2844


conf/console.default.ini
```diff
@@ -27,7 +27,10 @@ oidc.client.id =
 oidc.client.secret = 
 # This scope value is specific to your ident provider and is used in an token API request
 oidc.scope = urn/globus/auth/scope/transfer.api.globus.org/all
-# return URL hostname. This is required to be the hostname that this console runs as, so callback from ident provider can return control back to this application. If the console is installed on the UFS host (as it sometimes is), then this value would be the same as ufshost above. It may be different based on where the console is running.
+# oidc.console.hostnane is the return URL hostname. This is required to be the hostname that this console runs as,
```
dkavanagh: s/nane/name

eucaconsole/config.py
```diff
@@ -61,33 +61,14 @@
     __version__ = 'DEVELOPMENT'
 
 REQUIRED_CONFIG = {
-    # 'use': 'egg:eucaconsole',  # can't test for this since container strips it
     'pyramid.includes': ['pyramid_beaker', 'pyramid_chameleon', 'pyramid_layout'],
     'session.type': 'cookie',
-    'cache.memory': 'dogpile.cache.pylibmc'
+    'session.httponly': 'true',
```
dkavanagh: I don't think this is required as much is it's preferred. Some may be satisfied doing SSL termination at the ELB and trusting traffic inside the DC. Can we say for certain a customer hasn't turned this off? I wouldn't want to break that customer.

kamalgill: @dkavanagh : This has nothing to do with HTTPS, as it relates to HTTPOnly cookies.  See https://www.owasp.org/index.php/HTTPOnly

dkavanagh: Ah, thanks. I was confusing this with session.secure

kamalgill: Yeah, it has a confusing name -- all the more reason to move it out of the default console.ini

PR #2131, opened 2016-12-06
GUI-2823: Implement VPCs landing page

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2823

### DIFF
--- a/eucaconsole/forms/vpcs.py
+++ b/eucaconsole/forms/vpcs.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+# Copyright 2013-2016 Hewlett Packard Enterprise Development LP
+#
+# Redistribution and use of this software in source and binary forms,
+# with or without modification, are permitted provided that the following
+# conditions are met:
+#
+# Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# Redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+Forms for VPC resources
+
+"""
+from ..i18n import _
+from . import BaseSecureForm, ChoicesManager
+
+
+class VPCsFiltersForm(BaseSecureForm):
+    """Form class for filters on landing page"""
+
+    def __init__(self, request, ec2_conn=None, cloud_type='euca', **kwargs):
+        super(VPCsFiltersForm, self).__init__(request, **kwargs)
+        self.request = request
+        self.cloud_type = cloud_type
+        self.ec2_choices_manager = ChoicesManager(conn=ec2_conn)
+        self.facets = [
+            {'name': 'state', 'label': _('State'), 'options': self.get_state_choices()},
+            {'name': 'availability_zone', 'label': _('Availability zone'),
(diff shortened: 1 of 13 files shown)


eucaconsole/layout.py
```diff
@@ -71,9 +71,11 @@ def __init__(self, context, request):
         self.access_id = self.request.session.get('access_id')
         self.has_regions = True
         self.default_region = ''
+        self.is_vpc_supported = 'VPC' in request.session.get('supported_platforms', [])
```
dkavanagh: BaseView has a method with this logic. Could refs to that be replaced with refs to this? "request.layout_manager.layout.is_vpc_supported" would work.

eucaconsole/views/vpcs.py
```diff
@@ -28,16 +28,139 @@
...
+    @view_config(route_name='vpcs_json', renderer='json', request_method='POST')
+    def vpcs_json(self):
+        if not(self.is_csrf_valid()):
+            return JSONResponse(status=400, message="missing CSRF token")
+
+        with boto_error_handler(self.request):
+            vpc_items = self.vpc_conn.get_all_vpcs() if self.vpc_conn else []
+            subnets = self.vpc_conn.get_all_subnets()
+            route_tables = self.vpc_conn.get_all_route_tables()
+            internet_gateways = self.vpc_conn.get_all_internet_gateways()
```
dkavanagh: This could be passing filter values, right? If we're supporting server-side filtering, doing it on the other side of the API is better.

kamalgill: @dkavanagh: What filter values do you suggest passing?  Filters make more sense on the VPC details page AFAICT.

eucaconsole/forms/vpcs.py
```diff
@@ -0,0 +1,57 @@
...
+
+"""
+from ..i18n import _
+from . import BaseSecureForm, ChoicesManager
+
+
+class VPCsFiltersForm(BaseSecureForm):
+    """Form class for filters on landing page"""
+
+    def __init__(self, request, ec2_conn=None, cloud_type='euca', **kwargs):
```
kamalgill: @dkavanagh: The facets are built from form choices, which is consistent with how we've done this everywhere else.  No need to depart from that here IMO.

eucaconsole/layout.py
kamalgill: @dkavanagh: Possibly, I'll take a look...

eucaconsole/forms/vpcs.py
dkavanagh: The choices manager is used to supply the values, but that's the only value in the forms package when generating facets. I opted to create some facets right in the view (see metrics.py) as we're trying to move away from using wtforms, right? I hate to see more code being thrown in there when there's really no need.

eucaconsole/views/vpcs.py
dkavanagh: On the landing pages, we do all filtering on the client side, so all the methods below to "filter_X_by_Y()" aren't strictly necessary. But, if you do wish to provide filtering, use values in the API for that. For example, the get_all_subnets() call allows you to pass a vpc_id, therefore, filter_subnets_by_vpc() is completely un-necessary. Follow?

kamalgill: @dkavanagh: Fetching all subnets without filtering by vpc-id allows us to fetch the subnets for all VPCs via one API call rather than one call per VPC.  Ditto for route tables and internet gateways.

eucaconsole/forms/vpcs.py
kamalgill: @dkavanagh: Placing the facets in the form allows reuse of the get_options_from_choices() method (for the AZ choices), which is on BaseSecureForm, so moving that to the view is probably more trouble that it's worth at this point.  The metrics page built facets from static lists, which is a bit easier than the use case here.

eucaconsole/templates/vpcs/vpcs.pt
```diff
@@ -0,0 +1,137 @@
...
+            </div>
+            <div metal:fill-slot="tile_dropdown_button" tal:omit-tag="">
+                <a id="tile-item-dropdown_{{ item.name | escapeHTMLTagID }}" class="tiny secondary button dropdown right"
+                   data-dropdown="item-dropdown_{{ item.name | escapeHTMLTagID }}"><i class="grid-action"></i></a>
+                <ul id="item-dropdown_{{ item.name | escapeHTMLTagID }}" class="f-dropdown" data-dropdown-content="">
+                    <li><a i18n:translate="" ng-href="${prefix}/{{ item.name | escapeURL }}">View details</a></li>
+                </ul>
+            </div>
+            <div metal:fill-slot="tile_header" tal:omit-tag="">
+                <a ng-href="${prefix}/{{ item.name | escapeURL }}">{{ item.name }}</a>
```
kamalgill: @dkavanagh: Note the errant item.name references will be replaced with item.id, which I've already started doing in the GUI-2825 branch.

eucaconsole/layout.py
kamalgill: @dkavanagh: That seems beyond the scope of the ticket.  Perhaps open a new ticket?

kamalgill: @dkavanagh: referencing BaseView.is_vpc_supported could work via layout.py, as the former is a static method and we can pass the request object.  I'll give that a try.

kamalgill: Cool, referencing BaseView.is_vpc_supported(request) works fine in layout.py

eucaconsole/views/vpcs.py
dkavanagh: I was just considering complexity and we have 4 variables representing the number of vpcs, subnets, gateways and route tables (let's say V, S, G, R). The way the code is written, the iterations in the code can be captured by O(V*(S+G+R)). If one iterated though each resource, attaching values to the VPC by hash lookup, that complexity could be reduced to O(V+S+G+R). That's what I was talking about with the efficiently comment.  It all affects load time for the data.

kamalgill: @dkavanagh: I think I understand what you're suggesting here, but given an account has a default limit of 5 VPCs I wonder if it's worth pursuing?  I'll try it on the GUI-2825 WIP branch and run some benchmarks.

PR #2139, opened 2016-12-15
GUI-2825: Implement VPC details page

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2825

PR #2146, opened 2016-12-20
GUI-2824: Implement Create and Delete VPC

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2824

NOTE: #2139 should be merged prior to this PR, as GUI-2824 was branched off GUI-2825


eucaconsole/views/vpcs.py
```diff
@@ -326,9 +380,10 @@ def get_vpc_subnets(self):
             ))
         return subnets_list
 
-    def get_subnet_instances(self, subnet_id=None, vpc_reservations=None):
+    @staticmethod
+    def get_subnet_instances(subnet_id=None, vpc_reservations=None):
```
dkavanagh: I guess I'd suggest that these should not default to invalid values, right? This function will return quickly if either is None, so why even have default?

eucaconsole/views/vpcs.py
```diff
@@ -338,9 +393,10 @@ def get_subnet_instances(self, subnet_id=None, vpc_reservations=None):
                         ))
         return instances
 
-    def get_subnet_network_acls(self, subnet_id=None, vpc_network_acls=None):
+    @staticmethod
+    def get_subnet_network_acls(subnet_id=None, vpc_network_acls=None):
```
dkavanagh: same as above, no default needed here.

eucaconsole/views/vpcs.py
```diff
@@ -352,7 +408,7 @@ def get_subnet_network_acls(self, subnet_id=None, vpc_network_acls=None):
 
     def get_subnet_route_tables(self, subnet_id=None, vpc_route_tables=None):
```
dkavanagh: ditto

eucaconsole/views/vpcs.py
```diff
@@ -419,3 +469,59 @@ def update_internet_gateway(self):
...
+        if self.create_vpc_form.validate():
+            name = self.request.params.get('name')
+            cidr_block = self.request.params.get('cidr_block')
+            internet_gateway = self.request.params.get('internet_gateway')
+            tags_json = self.request.params.get('tags')
+            self.log_request(_(u"Creating VPC {0}").format(name))
+            with boto_error_handler(self.request):
+                new_vpc = self.vpc_conn.create_vpc(cidr_block)
+                new_vpc_id = new_vpc.id
+                self.log_request(_(u"Created VPC {0} ({1})").format(name, new_vpc_id))
```
dkavanagh: We typically log before the operation and then any failure will show after using the correlation ID. That way the order in logging is consistent.

eucaconsole/templates/vpcs/vpc_view.pt
```diff
@@ -209,6 +212,85 @@
...
+                            </div>
+                            <div class="columns small-9">${vpc_internet_gateway_name}</div>
+                        </div>
+                        <div tal:condition="vpc_route_tables" class="row">
+                            <div class="columns small-3" i18n:translate="">
+                                Route tables:
+                            </div>
+                            <div class="columns small-9">
+                                <span tal:repeat="route_table vpc_route_tables">
+                                    ${layout.get_display_name(route_table)}<span
```
dkavanagh: Curious why you decided to invoke this from template where we normally throw the display name in the render dict in the view?

eucaconsole/forms/vpcs.py
```diff
@@ -40,6 +40,9 @@
     'An internet gateway must be attached to a VPC if you wish to create one or more public subnets in the VPC.'
 )
 
+CIDR_BLOCK_REGEX = '^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}' \
+                    '([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/([0-9]|[1-2][0-9]|3[0-2]))$'
+
```
kamalgill: Good catch -- updated the one in layout.py to reuse the new (and more accurate) regex.

eucaconsole/templates/vpcs/vpc_new.pt
```diff
@@ -0,0 +1,70 @@
...
+        </metal:breadcrumbs>
+        <!-- Notifications -->
+        <metal:block metal:use-macro="layout.global_macros['notifications']" />
+        <h3 id="pagetitle">
+            <strong i18n:translate="">Create VPC</strong>
+        </h3>
+        <div class="large-12 columns">
+            <div class="panel">
+                <form action="${request.route_path('vpc_create')}" method="post"
+                      data-abide="abide" id="create-vpc-form">
```
kamalgill: I had trouble with Angular validation (e.g. issues with form's pristine state), so I had to fall back to Abide validation.  We can probably revisit this later, perhaps in a separate ticket.

eucaconsole/templates/vpcs/vpc_new.pt
```diff
@@ -0,0 +1,70 @@
...
+                            <metal:block metal:use-macro="layout.global_macros['required_fields_label']" />
+                        </div>
+                    </div>
+                </form>
+            </div>
+        </div>
+        <div class="large-12 columns">
+            <metal:block metal:use-macro="layout.global_macros['help_expando']" />
+            <div class="help-content" ng-show="isHelpExpanded" ng-cloak="">
+                <!--! Help content here -->
```
kamalgill: Thanks for the reminder -- I'll coordinate with Keri to ensure we have DOC tickets available for the VPC pages as we get closer to being feature-complete.

eucaconsole/templates/vpcs/vpc_view.pt
kamalgill: We need the helper method here due to the tal:repeat.

dkavanagh: Here's an example of how we've done this before, which is why I asked about the difference: https://github.com/eucalyptus/eucaconsole/blob/master/eucaconsole/views/instances.py#L526

kamalgill: @dkavanagh: Ah, understood.  I prefer we use the layout helper method instead of adding an attribute to the objects though.

eucaconsole/templates/vpcs/vpc_new.pt
dkavanagh: Maybe we should talk. Did you try "novalidate" as an form attribute? That turns off the browser validation and lets angular handle it.

eucaconsole/templates/vpcs/vpc_view.pt
dkavanagh: I'll also accept "since we're not iterating through that list in the view method, we can use the tal:repeat iteration to accomplish this" :-)

PR #2148, opened 2016-12-22
GUI-2851 filled out preferences tab content. superficially functional…

dkavanagh:
…, mostly nothing behind it

https://eucalyptus.atlassian.net/browse/GUI-2851


eucaconsole/templates/reporting/reporting_preferences.pt
```diff
@@ -1,3 +1,89 @@
 <div>
```
dkavanagh: Added missing i18n:domain attr in GUI-2852 branch

PR #2152, opened 2017-01-11
GUI-2827: Implement Subnet Details page

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2827

### DIFF
--- a/eucaconsole/forms/__init__.py
+++ b/eucaconsole/forms/__init__.py
@@ -474,7 +474,8 @@ class ChoicesManager(object):
         from ..views import TaggedItemView
         choices = []
         if add_blank:
-            choices = [('None', _(u'None'))]
+            # NOTE: Do not mark the 'None' value for i18n
+            choices = [('None', _(u'None (use main route table for VPC)'))]
         filters = {}
         if vpc is not None:
             filters.update({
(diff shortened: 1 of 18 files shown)


eucaconsole/forms/__init__.py
```diff
@@ -474,7 +474,8 @@ def vpc_route_tables(self, vpc=None, route_tables=None, add_blank=True, escapebr
         from ..views import TaggedItemView
         choices = []
         if add_blank:
-            choices = [('None', _(u'None'))]
+            # NOTE: Do not mark the 'None' value for i18n
+            choices = [('None', _(u'None (use main route table for VPC)'))]
```
kamalgill: See the update_route_table() method in SubnetView re how we're handling this.

dkavanagh: ic. In many other places we use the empty string for this case, which is like "no selection", right?

kamalgill: We've used 'None' in other VPC resources, security groups, and key pairs, and 'None' is more explicit than '', so I stuck with that here.  I see your point about using an empty string, but we've typically used that only for the "Select..." option, which won't usually be passed across the wire due to client-side validation.

dkavanagh: That's fine, just trying to understand the usage.

PR #2154, opened 2017-01-11
GUI-2857 added support for user report access enablement flag

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2857

### DIFF
--- a/eucaconsole/views/reporting.py
+++ b/eucaconsole/views/reporting.py
@@ -68,11 +68,11 @@ class ReportingAPIView(BaseView):
         ret = self.conn.view_account()
         acct_prefs = ret.get('accountSettings')
         ret = dict(
-            enabled=True if prefs.get('detailedBillingEnabled') else False,
+            enabled=prefs.get('detailedBillingEnabled'),
             bucketName=prefs.get('reportBucket') or '',
             activeTags=prefs.get('activeCostAllocationTags') or ['one'],
             inactiveTags=prefs.get('inactiveCostAllocationTags') or ['two', 'three', 'four'],
-            userReportsEnabled=True if acct_prefs.get('userBillingAccess') else False,
+            userReportsEnabled=acct_prefs.get('userBillingAccess'),
         )
         return dict(results=ret)
 
--- a/eucaconsole/views/reporting.py
+++ b/eucaconsole/views/reporting.py
@@ -33,7 +33,6 @@ import simplejson as json
 from pyramid.view import view_config
 
 from ..i18n import _
-from ..views import BaseView
 from ..views import BaseView, JSONResponse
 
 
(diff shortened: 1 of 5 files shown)


eucaconsole/views/reporting.py
```diff
@@ -65,11 +65,14 @@ def get_reporting_prefs(self):
         # use "ViewBilling" call to fetch billing configuration information
         ret = self.conn.view_billing()
         prefs = ret.get('billingSettings')
+        ret = self.conn.view_account()
```
kamalgill: We should probably use a more descriptive name for this variable rather than overriding the one two lines up.

eucaconsole/views/reporting.py
```diff
@@ -65,11 +65,14 @@ def get_reporting_prefs(self):
...
+        acct_prefs = ret.get('accountSettings')
         ret = dict(
             enabled=True if prefs.get('detailedBillingEnabled') else False,
             bucketName=prefs.get('reportBucket') or '',
             activeTags=prefs.get('activeCostAllocationTags') or ['one'],
-            inactiveTags=prefs.get('inactiveCostAllocationTags') or ['two', 'three', 'four']
+            inactiveTags=prefs.get('inactiveCostAllocationTags') or ['two', 'three', 'four'],
+            userReportsEnabled=True if acct_prefs.get('userBillingAccess') else False,
```
kamalgill: Minor point, but this line can be simplified to `userReportsEnabled=bool(acct_prefs.get('userBillingAccess'))`.  Fine to leave it though, since it's fairly clear as-is.

eucaconsole/views/reporting.py
```diff
@@ -79,10 +82,14 @@ def set_reporting_prefs(self):
         csrf_token = params.get('csrf_token')
         if not self.is_csrf_valid(token=csrf_token):
             return JSONResponse(status=400, message="missing CSRF token")
+        self.log_request(_(u"Saving report preferences"))
```
kamalgill: Is a generic log statement covering the modify_billing() and modify_account() actions adequate, or should we log those separately?

eucaconsole/views/reporting.py
dkavanagh: It's just a temp variable, like in the usage above. Could call them both tmp, or result, but it's all the same.

eucaconsole/views/reporting.py
dkavanagh: bool('true')

eucaconsole/views/reporting.py
dkavanagh: It's not about logging api actions, the back end can do that. This is about logging things the user is doing. So, that is why we tend to log what they've requested to do and we're doing now. Like when we create an ELB, but may make several API calls to get that done.

PR #2159, opened 2017-01-19
Gui 2822 fixed validation error when selecting instances first

dkavanagh:
This also links new wizard into navigation.

https://eucalyptus.atlassian.net/browse/GUI-2822


eucaconsole/static/js/services/elb-service.js
```diff
@@ -12,14 +12,14 @@ angular.module('ELBServiceModule', [])
         createELB: function (csrfToken, values) {
             // update listeners to be passed to API call
             var listeners = JSON.parse(JSON.stringify(values.listeners));
```
kamalgill: Why do we need the JSON stringify-and-parse approach here?

dkavanagh: I wish I could remember. I should have added a comment about that.

PR #2161, opened 2017-01-23
Gui 2864 Reports landing template

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2864

### DIFF
--- a/eucaconsole/static/css/pages/reporting.css
+++ b/eucaconsole/static/css/pages/reporting.css
@@ -6,6 +6,163 @@
 /* Set the link colors and styles for top-level nav */
 /* ----------------------------------------- Item list
 ----------------------------------------- */
+/* @section Landing pages styles */
+.columns.datagrid { padding-left: 0; padding-right: 1rem; padding-bottom: 2rem; }
+.columns.datagrid.has-smart-table.tableview #sorting-controls { display: none; }
+
+.large-12.columns.datagrid { padding-left: 1rem; }
+
+.clear-link { color: #666666; }
+.clear-link:hover { color: #01a982; }
+
+#breadcrumbs .icon-block { display: none; }
+
+#primary-buttons .button { padding: 8px 1rem; }
+#primary-buttons .button.more-actions { margin-left: 1rem; padding-right: 3rem; }
+#primary-buttons .button.more-actions[disabled] { border: 2px solid #cccccc; }
+#primary-buttons .button.more-actions[disabled]:hover { outline: none; }
+#primary-buttons .button.more-actions[disabled]:after { border-color: #cccccc transparent transparent transparent; }
+
+#more-actions-dropdown.f-dropdown { border-color: #cccccc; box-shadow: 2px 2px 5px #F0F0F0; }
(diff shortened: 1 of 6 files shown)


eucaconsole/templates/reporting/reporting_reports.pt
```diff
@@ -1,3 +1,61 @@
-<div>
-<span>Reports</span>
+<div i18n:domain="eucaconsole">
+    <div class="section">
+        <h6 class="section-label" i18n:translate="">Usage Reports</h6>
+        <a i18n:translate="" ng-click="reports.showEC2InstanceUsageReport()">EC2 Instance Usage Report</a>
```
kamalgill: The EC2 Instance Usage Report (and the Usage Reports by Service) should link to external pages.

dkavanagh: That's the sort of thing we'll change as we code the ec2 instance usage reports.

eucaconsole/static/css/pages/reporting.css
```diff
@@ -6,6 +6,163 @@
 /* Set the link colors and styles for top-level nav */
 /* ----------------------------------------- Item list
 ----------------------------------------- */
+/* @section Landing pages styles */
```
kamalgill: There's probably no need to pull in the landing page CSS, despite the reports landing page name.

dkavanagh: This was to get table styling. Maybe table styling an be pulled out separately? It was needed for stack events table and imported the same way.

kamalgill: The table on the reports landing page doesn't really need the styles of the resource landing pages, esp. since it doesn't need sortable headings, filters, MSB, etc.  Foundation's vanilla table styles should be good enough here.

dkavanagh: Thinking of consistency here, right?

kamalgill: Ah, I forgot we created the table-styles-mixin for this purpose.  See https://github.com/eucalyptus/eucaconsole/blob/master/eucaconsole/static/sass/includes/_eucavariables.scss#L294

aliceh:
retest this please

eucaconsole/static/css/pages/reporting.css
dkavanagh: We included that both in landing pages and eucaconsole, so not sure why I wasn't getting those styles. I'll look into it.
eucaconsole/static/sass//eucaconsole.scss:    @include table-styles-mixin;
eucaconsole/static/sass//includes/_landingpage.scss:    @include table-styles-mixin;

dkavanagh: The difference.. eucaconsole uses "table.table" selector and does not set width to 100%, landingpage uses ".gridwrapper .table" selector and does set width to 100%. I could simply add that missing attribute in this page's sass

dkavanagh:
closing because another PR encompasses these changes.

PR #2162, opened 2017-01-23
GUI-2826: Implement Add Subnet feature on VPC details page

kamalgill:
https://eucalyptus.atlassian.net/browse/GUI-2826


aliceh:
retest this please

eucaconsole/forms/vpcs.py
```diff
@@ -161,6 +161,28 @@ def __init__(self, request, vpc_conn=None, **kwargs):
...
+        validators=[
+            validators.InputRequired(message=cidr_block_error_msg),
+            validators.Regexp(CIDR_BLOCK_REGEX, message=cidr_block_error_msg),
+        ]
+    )
+    availability_zone = SelectField(label=_('Availability zone'))
+
+    def __init__(self, request, ec2_conn=None, **kwargs):
+        super(CreateSubnetForm, self).__init__(request, **kwargs)
+        self.ec2_choices_manager = ChoicesManager(conn=ec2_conn)
```
dkavanagh: not a big deal, but when the value isn't used outside this function, keep it local, right?

eucaconsole/views/vpcs.py
```diff
@@ -311,6 +314,24 @@ def vpc_delete(self):
             self.request.error_messages = self.vpc_delete_form.get_errors_list()
             return self.render_dict
 
+    @view_config(route_name='vpc_add_subnet', renderer=VIEW_TEMPLATE, request_method='POST')
+    def vpc_add_subnet(self):
+        location = self.request.route_path('vpc_view', id=self.vpc.id)
+        if self.add_subnet_form.validate():
+            name = self.request.params.get('subnet_name')
+            cidr_block = self.request.params.get('subnet_cidr_block')
+            zone = self.request.params.get('availability_zone')
```
dkavanagh: Here, we should log that we're creating the resource.

eucaconsole/templates/vpcs/vpc_view.pt
```diff
@@ -179,6 +180,45 @@
...
+                        Use the CIDR format to specify your subnet's IP address block (e.g. 10.0.0.0/24).
+                        Note that block sizes must be between a /16 netmask and a /28 netmask.
+                        Also, note that a subnet can be the same size as your VPC.
+                    </span>
+                </p>
+                <div>
+                    ${panel('form_field', field=add_subnet_form['subnet_name'], stack_label=True)}
+                </div>
+                <div>
+                    ${panel('form_field', field=add_subnet_form['subnet_cidr_block'], stack_label=True, pattern=add_subnet_form.cidr_block_regex)}
```
dkavanagh: More of a UX comment, but I'll leave it here. Since the CIDR block is going to start the same as the ${vpc.cidr_block}, we could pre-populate, at least part, if not all of the block to help the user.

eucaconsole/views/vpcs.py
```diff
@@ -427,6 +430,36 @@ def get_subnet_network_acls(subnet_id, vpc_network_acls):
...
 
+    @staticmethod
+    def get_suggested_subnet_cidr_block(vpc_cidr_block):
+        """Suggest a subnet CIDR block based on the VPC's CIDR block"""
+        vpc_cidr_split = vpc_cidr_block.split('/')
+        cidr_ip = vpc_cidr_split[0]
+        cidr_netmask = int(vpc_cidr_split[1])
+        cidr_ip_parts = cidr_ip.split('.')
+        num_ip_parts = 2
+        ip_suffix = '.128.0'  # The 128 in '.128.0' is somewhat arbitrary but seems safer than suggesting '.0.0'
```
dkavanagh: Not sure that safety has much to do with it. I hadn't thought of this before, but when we suggest device name for attaching volumes, we look at what's there already. Could we do same here? Obviously much more complicated though. My initial thought when suggesting we pre-populate was that we at least can fill in the 1st part of the CIDR block based on the main CIDR def, i.e for 172.31.0.0/16, we could be 100% certain that "172.31." would be required. Without being completely sure the suggestion doesn't overlap an existing subnet, I'm hesitant to provide that complete a suggestion. We don't want to pre-fill with something that could fail.

eucaconsole/templates/vpcs/vpc_view.pt
```diff
@@ -194,8 +194,10 @@
                     <strong>${vpc.cidr_block}</strong>.
                     <span i18n:translate="">
                         Use the CIDR format to specify your subnet's IP address block (e.g. 10.0.0.0/24).
-                        Note that block sizes must be between a /16 netmask and a /28 netmask.
-                        Also, note that a subnet can be the same size as your VPC.
+                        Block sizes must be between a /16 netmask and a /28 netmask.
+                        A subnet can be the same size as your VPC if the subnet is the only one in the VPC.
+                        Note the pre-populated subnet CIDR block will need to be modified if it conflicts
+                        with an existing subnet's CIDR block in the VPC.
```
dkavanagh: If we do or don't pre-fill, we don't give the user enough information on this screen to decide. They must already know their network allocation. right?

kamalgill: I'm not sure what you mean by "we don't give the user enough information on this screen to decide."  The existing subnet's are displayed on the VPC details page, and the Add subnet modal displays the VPC's CIDR block.  What more should we add here?

eucaconsole/views/vpcs.py
kamalgill: I'd prefer to pre-fill the entire CIDR block than leave an invalid one, as '172.31.' is far more likely to fail than '172.31.128.0/20'.  Since the existing subnet CIDR blocks are displayed on the VPC details page, it isn't much to ask one to increment the IP portion if necessary IMO.

eucaconsole/templates/vpcs/vpc_view.pt
dkavanagh: When the user clicks "add subnet", the create subnet dialog covers existing subnet CIDRs, so while they do have the network's CIDR, they aren't able to see the existing subnet CIDR ranges, so the information at that time will be incomplete. The user would need to have a subnet CIDR in mind when clicking "Add Subnet".  I think the only way to improve that is by saying "here's the network CIDR, and existing subnet CIDRs, enter the new range.". 
I honestly don't know if people will be doing this without a subnet CIDR in mind. Maybe I'm making more out of this.

eucaconsole/views/vpcs.py
dkavanagh: I struggled with this as well. Try leaning towards not pre-populating with something that can fail. It's not good UX.

eucaconsole/templates/vpcs/vpc_view.pt
kamalgill: Listing the existing subnet CIDR blocks in the dialog is a good suggestion -- will make that change shortly.

PR #2164, opened 2017-01-31
GUI-2839: Allow broken cookie to be cleared (following console restart) when GovCloud region is mistakenly enabled

kamalgill:
Also, prevent duplicate "US GovCloud" entries from displaying in region choices.

https://eucalyptus.atlassian.net/browse/GUI-2839

### DIFF
--- a/eucaconsole/layout.py
+++ b/eucaconsole/layout.py
@@ -72,15 +72,10 @@ class MasterLayout(object):
         self.has_regions = True
         self.default_region = ''
         if self.cloud_type == 'aws':
-            self.regions = AWS_REGIONS
-            govcloud_region_option = dict(
-                name='us-gov-west-1',
-                label='US GovCloud',
-            )
-            if asbool(request.registry.settings.get('aws.govcloud.enabled', 'false')):
-                if govcloud_region_option not in self.regions:
-                    self.regions.append(govcloud_region_option)
             self.default_region = request.registry.settings.get('aws.default.region', 'us-east-1')
+            self.regions = list(AWS_REGIONS)
+            if asbool(request.registry.settings.get('aws.govcloud.enabled', 'false')):
+                self.regions.append(dict(name='us-gov-west-1', label='US GovCloud'))
         else:
             if self.access_id:
                 host = self.request.registry.settings.get('ufshost')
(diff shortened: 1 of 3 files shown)


eucaconsole/views/dashboard.py
```diff
@@ -69,6 +71,11 @@ def __init__(self, request):
 
     @view_config(route_name='dashboard', request_method='GET', renderer='../templates/dashboard.pt')
     def dashboard_home(self):
+        if self.conn is None:
+            # Handle broken connection case (e.g. configured AWS GovCloud region without proper access)
+            forget(self.request)
```
dkavanagh: I'm not sure you're using forget() as intended: http://docs.pylonsproject.org/projects/pyramid/en/latest/api/security.html#pyramid.security.forget

eucaconsole/layout.py
```diff
@@ -73,11 +73,13 @@ def __init__(self, context, request):
         self.default_region = ''
         if self.cloud_type == 'aws':
             self.regions = AWS_REGIONS
+            govcloud_region_option = dict(
+                name='us-gov-west-1',
+                label='US GovCloud',
+            )
             if asbool(request.registry.settings.get('aws.govcloud.enabled', 'false')):
```
dkavanagh: Curious why this change? It creates the dict in all cases as compared to the old code that only created it when needed.

eucaconsole/views/dashboard.py
dkavanagh: re-reading this and I noticed the additional condition. Was the gov cloud being appended multiple times?

kamalgill: Yes, the GovCloud label was appended multiple times.  Regarding the forget() method, we need that to clear the cookie to avoid a persistent internal server error.

eucaconsole/layout.py
kamalgill: Good catch, I'll move the dict inside the aws.govcloud.enabled check.

eucaconsole/views/dashboard.py
dkavanagh: OK, It wasn't clear to me that the forget method had that side effect.

PR #2165, opened 2017-02-01
GUI-2829: Implement route table details page

kamalgill:
Implements https://eucalyptus.atlassian.net/browse/GUI-2829

NOTE: This is a work in progress -- more work is needed in GUI-2829 to incorporate NAT gateways as route targets


eucaconsole/forms/vpcs.py
```diff
@@ -233,3 +233,26 @@ def set_choices(self):
...
+    """Form to update an existing route table"""
+    name_error_msg = _('Not a valid name')
+    name = TextEscapedField(label=_('Name'))
+
+    def __init__(self, request, route_table=None, **kwargs):
+        super(RouteTableForm, self).__init__(request, **kwargs)
+        self.route_table = route_table
+        self.name.error_msg = self.name_error_msg
+
+        if self.route_table:
```
dkavanagh: Will route table ever be None? I notice one is always passed to the form in the view.

eucaconsole/templates/vpcs/route-table-editor.pt
```diff
@@ -0,0 +1,70 @@
...
+                   ng-pattern="cidrPattern" />
+        </div>
+        <div class="small-6 columns">
+            <label for="route-target" i18n:translate="">
+                Target
+            </label>
+            <select ng-model="routeTarget" name="routeTarget" id="route-target" ng-disabled="routeTargetsLoading">
+                <option value="" i18n:translate="" ng-if="routeTargetsLoading">Loading...</option>
+                <option value="" i18n:translate="" ng-if="!routeTargetsLoading">Select...</option>
+                <option ng-repeat="targetChoice in routeTargets" value="{{ targetChoice.id }}">
```
dkavanagh: why not use ng-options?

eucaconsole/templates/vpcs/route_table_view.pt
```diff
@@ -0,0 +1,165 @@
...
+                                    ${layout.get_display_name(subnet)}
+                                </a>
+                            </div>
+                        </div>
+                    </div>
+                    <hr />
+                    <h6 i18n:translate="">Routes</h6>
+                    <div ng-cloak="">
+                        <route-table-editor ng-model="routes" cidr-pattern="${layout.cidr_pattern}" vpc-id="${vpc.id}"
+                                            tal:content="routes | structure">
```
dkavanagh: what does this do, considering this directive has a template that determines the content?

eucaconsole/templates/vpcs/route_table_view.pt
```diff
@@ -0,0 +1,165 @@
...
+                        </route-table-editor>
+                    </div>
+                    <hr />
+                    <div ng-cloak="">
+                        <tag-editor ng-model="tags" show-name-tag="false" tal:content="tags | structure"></tag-editor>
```
dkavanagh: same question here about tal:content

eucaconsole/views/vpcs.py
```diff
@@ -738,3 +743,206 @@ def get_controller_options_json(self):
...
+
+    def __init__(self, request, **kwargs):
+        super(RouteTableView, self).__init__(request, **kwargs)
+        self.location = self.request.route_path(
+            'route_table_view', vpc_id=self.request.matchdict.get('vpc_id'), id=self.request.matchdict.get('id'))
+        self.conn = self.get_connection()
+        self.vpc_conn = self.get_connection(conn_type='vpc')
+        with boto_error_handler(request, self.location):
+            self.vpc = self.get_vpc()
+            self.route_table = self.get_route_table()
```
dkavanagh: Since route table really required for this view, if you fail early, can avoid several checks for it later.

eucaconsole/templates/vpcs/route_table_view.pt
kamalgill: This is how we've pulled in the new Angular-based tag editor everywhere else.  What change are you suggesting?

eucaconsole/templates/vpcs/route_table_view.pt
kamalgill: I followed the pattern used in the tag editor directive here.

eucaconsole/forms/vpcs.py
kamalgill: Probably not -- I'll remove this check.

eucaconsole/templates/vpcs/route-table-editor.pt
kamalgill: ng-repeat easily allows the Loading/Select option above the choices, and ng-repeat is more readable than ng-options IMO.

eucaconsole/templates/vpcs/route_table_view.pt
dkavanagh: Technically, it's inserting "routes" in as structure, but doesn't the template in the directive replace that? Try removing tal:content.

eucaconsole/views/vpcs.py
kamalgill: What are you suggesting here?  self.vpc is also required for this view.

eucaconsole/templates/vpcs/route_table_view.pt
kamalgill: See the first two lines of the directive's link() method on why we need the tal:content.  Ditto for the tag editor.

dkavanagh: OK. I forgot about transclusion.

eucaconsole/routes.py
```diff
@@ -279,6 +279,11 @@ def instances_landing(self):
     Route(name='subnet_view', pattern='/vpcs/{vpc_id}/subnets/{id}'),
     Route(name='subnet_update', pattern='/vpcs/{vpc_id}/subnets/{id}/update'),
     Route(name='subnet_delete', pattern='/vpcs/{vpc_id}/subnets/{id}/delete'),
+    Route(name='route_table_view', pattern='/vpcs/{vpc_id}/route-tables/{id}'),
+    Route(name='route_table_update', pattern='/vpcs/{vpc_id}/route-tables/{id}/update'),
+    Route(name='route_table_delete', pattern='/vpcs/{vpc_id}/route-tables/{id}/delete'),
```
dkavanagh: weren't we going to a more restful model where the http method indicated the action more than the url?

eucaconsole/static/js/services/route-target-service.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright 2017 Hewlett Packard Enterprise Development LP
+ *
+ * @fileOverview Route target service module
+ * @requires AngularJS
+ *
+ */
+angular.module('RouteTargetServiceModule', [])
```
dkavanagh: fwiw, when something is this specific and used in only one page, the service can be pulled into the module for that page.

kamalgill: this is going to be used for the Create Route Table page as well

eucaconsole/routes.py
kamalgill: we're using a simple form post for the delete

eucaconsole/views/vpcs.py
dkavanagh: If VPC and route table are required and they aren't found, we should be raising a 404, right?

dkavanagh: I see the not found in the view method. Hmm. probably the only place we can do that. Just too bad this other common code needs conditionals before that can be raised.

dkavanagh: Actually, I just tested this and the __init__ can raise HTTPNotFound. If this view is only used for existing route tables, that could be cleaner. fails faster, blah, blah.

PR #2166, opened 2017-02-02
Gui 2866 fill out services for monthly usage report, finish client side styling

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2866


eucaconsole/static/sass/pages/reporting.scss
```diff
@@ -1,4 +1,5 @@
 @import "../includes/eucavariables";
+@import "../includes/landingpage";
```
kamalgill: Let's use the table styles mixin instead of pulling in the entire landing page CSS.

eucaconsole/views/reporting.py
```diff
@@ -93,3 +111,54 @@ def set_reporting_prefs(self):
...
+                results.append((rec[0],))
+                service = rec[0]
+            results.append((
+                rec[0], rec[1],
+                '{:0.8f}'.format(rec[2]).rstrip('0').rstrip('.'),
+                self.units_from_details(rec[1])
+            ))
+        return dict(results=results)
+
+    @view_config(route_name='reporting_monthly_usage', request_method='POST')
```
kamalgill: Why is the CSV download a POST request, while the JSON fetch is a GET?  They both should be GET requests, right?

eucaconsole/static/js/pages/reporting_reports.js
```diff
@@ -0,0 +1,34 @@
+angular.module('ReportingPage')
+.controller('ReportsController', ['$scope', '$routeParams', 'ReportingService', 'eucaHandleError', function ($scope, $routeParams, ReportingService, eucaHandleError) {
```
kamalgill: We should add a line break here to control line length and facilitate code reviews.

eucaconsole/static/js/pages/reporting.js
```diff
@@ -38,10 +38,6 @@ angular.module('ReportingPage', ['ngRoute', 'localytics.directives', 'EucaConsol
     console.log('dashboard');
     // would be nice to keep this controller function in a separate file to avoid this getting bloated
 }])
-.controller('ReportsController', ['$scope', '$routeParams', function ($scope, $routeParams) {
-    console.log('reports');
-    // would be nice to keep this controller function in a separate file to avoid this getting bloated
-}])
 .config(function ($routeProvider, $locationProvider) {
```
kamalgill: Combining the reporting tabs into a single page is a slipper slope IMO.  As we previously discussed, the benefits do not outweigh the drawbacks.

eucaconsole/views/reporting.py
dkavanagh: I was using POST because that's what the "generateFile" module uses.

eucaconsole/static/sass/pages/reporting.scss
dkavanagh: Remember in the other PR, that mixin is included in both places. The special sauce is in the landing pages css. I could simply duplicate the part I need again.

eucaconsole/static/js/pages/reporting.js
dkavanagh: Yes. After using this, it's even more apparent why single page is nicer for UX. When the user navigates to "usage reports", they're taken to the dashboard. I've been testing the reporting tab, so nav to that is a split second thing by clicking that tab. It's pretty nice to use.

kamalgill: Please help me understand how fast tab-switching to an unrelated page is better UX than significantly increasing the initial page load time (by combining those unrelated pages into a single page).

dkavanagh: First, they are very clearly related as being all about reports and usage. Second, the navigation item (in the user menu) links to the main reports page (dashboard). To see monthly usage, I need to click the "Reports" tab after that. I'm pointing out that there's less friction in that when the user doesn't have to wait for a page load. Have you tried using the pages that are there? How does it "feel" to you?

PR #2170, opened 2017-02-03
GUI-2875: Fix policy editor textarea position to keep it visible when page scrolls

kamalgill:
Fixes https://eucalyptus.atlassian.net/browse/GUI-2875 (for 4.4)

### DIFF
--- a/eucaconsole/static/sass/pages/iam_policy_wizard.scss
+++ b/eucaconsole/static/sass/pages/iam_policy_wizard.scss
@@ -57,6 +57,14 @@
     }
 }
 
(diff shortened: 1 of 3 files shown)


eucaconsole/static/sass/pages/iam_policy_wizard.scss
```diff
@@ -57,6 +57,14 @@
     }
 }
 
+@media only screen and (min-width: 64.063em) {  // 64.063em = large screen width
```
dkavanagh: I notice we're somewhat inconsistent with our breakpoint definitions, some in eucavariables, several not.

PR #2176, opened 2017-02-10
GUI-2869:  Add executables and systemd services

gholms:
We currently maintain one long rc script that does five things:  session key generation, certificate generation, managing memcached, managing nginx, and managing eucaconsole itself.  Systemd's service dependency management allows us to replace that code with a service file for each of these functions, while retaining the same behavior for end users that we have now.

This patch set makes the following changes to the code base:
- Service-related files land in a new `systemd` directory.  Rather than installing them unconditionally, `setup.py install` only installs them when given a `--init-system=systemd` option to prevent excess cruft from being installed on systems that do not use systemd.
- The `eucaconsole` main executable that never made it out of the packaging code repository for whatever reason lands in a new `bin` directory.  This program accepts two options:  `-c`, which specifies the location of a configuration file, and `-r`, which enables automatic restarting during development similar to what `launcher.sh` does.
- The session key and certificate generation code that was previously part of the rc script lands in two scripts:  `bin/eucaconsole-keygen` and `bin/eucaconsole-certgen`.  These scripts are idempotent, installed in `/usr/libexec/eucaconsole`, and called by systemd when the files they respectively generate are missing or empty.
- `setup.py install` installs `console.ini` itself.  The file in the source tree is renamed from `console.default.ini` because making distutils rename a data file as it installs it is difficult.
- The `COPYING` file is included in source distributions.

Fixes [GUI-2869](https://eucalyptus.atlassian.net/browse/GUI-2869)


aliceh:
Retest this please

gholms:
Rest this please.  It looks like a network failure of some sort.

gholms:
Retest this please.  It looks like a network failure of some sort.

bin/eucaconsole
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/python
```
dkavanagh: We used the -tt option to get more diagnostic output captured in the eucaconsole_startup.log

dkavanagh:
Nice to be able to ditch the daemon overhead.

PR #2177, opened 2017-02-13
GUI-2870, GUI-2871: Create/View NAT Gateway

kamalgill:
Implements:
- https://eucalyptus.atlassian.net/browse/GUI-2870
- https://eucalyptus.atlassian.net/browse/GUI-2871

Fixes:
- https://eucalyptus.atlassian.net/browse/GUI-2868

### DIFF
--- a/eucaconsole/views/__init__.py
+++ b/eucaconsole/views/__init__.py
@@ -119,7 +119,7 @@ class BaseView(object):
         self.euca_logout_form = EucaLogoutForm(self.request)
 
     def get_connection3(self, conn_type='ec2', cloud_type=None, region=None, access_key=None,
-                       secret_key=None, security_token=None):
+                        secret_key=None, security_token=None):
         # For this spike, rely on existing model/auth.py code to do the hard stuff.
         # later, we'd convert all that from the ground up
         conn2 = self.get_connection(conn_type, cloud_type, region, access_key, secret_key, security_token)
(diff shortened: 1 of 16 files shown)


eucaconsole/views/__init__.py
```diff
@@ -128,18 +128,24 @@ def get_connection3(self, conn_type='ec2', cloud_type=None, region=None, access_
...
             aws_session_token=conn2.provider.security_token,
-            api_version=conn2.APIVersion,
             use_ssl=conn2.is_secure,
-            endpoint_url=endpoint_url,
             verify=False
         )
+        if self.cloud_type == 'euca':
+            client_args.update(dict(
+                api_version=conn2.APIVersion,
+                endpoint_url=endpoint_url,
```
dkavanagh: If this only set for euca, how do we connect to various aws regions?

kamalgill: We're passing the region to session.create_client, as below...
```
conn3 = session.create_client(
    conn_type, conn2.region.name, **client_args
)
```

dkavanagh: ok, makes sense. Just need endpoint_url in cases where euca isn't supporting DNS names.

eucaconsole/views/vpcs.py
```diff
@@ -716,6 +725,29 @@ def subnet_delete(self):
...
+                    prefix = _(u'Successfully allocated elastic IPs')
+                    ips = ', '.join(new_ips)
+                    msg = u'{prefix} {ips}'.format(prefix=prefix, ips=ips)
+                    self.request.session.flash(msg, queue=Notification.SUCCESS)
+            self.request.session.flash(msg, queue=Notification.SUCCESS)
+            return HTTPFound(location=location)
+        else:
+            self.request.error_messages = self.allocate_eips_form.get_errors_list()
+        return self.render_dict
+
```
dkavanagh: It's too bad we couldn't have a single route path for ip address allocation since it's also handle din the IPAddressesView. Much code duplicated.

PR #2179, opened 2017-02-14
addendum to GUI-2869

mbacchi:


euca-qa-gui-tester:

PR #2184, opened 2017-03-06
GUI-2854: Support bucket policy configuration

kamalgill:
https://eucalyptus.atlassian.net/browse/GUI-2854


eucaconsole/static/js/pages/bucket_details.js
```diff
@@ -155,7 +220,10 @@ angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils', 'Cors
                         $scope.savingCorsConfig = true;
                         $scope.corsError = '';
                         var csrfToken = angular.element('#csrf_token').val();
-                        CorsService.setCorsConfig($scope.bucketName, csrfToken, $scope.codeEditor.getValue())
+                        var corsTextarea = angular.element('#cors-textarea');
+                        // Fall back to standard textarea in cases where CodeMirror fails to initialize
```
dkavanagh: Is this likely and how should we address this elsewhere?

eucaconsole/static/js/pages/bucket_details.js
```diff
@@ -166,6 +234,7 @@ angular.module('BucketDetailsPage', ['S3SharingPanel', 'EucaConsoleUtils', 'Cors
                                 $scope.savingCorsConfig = false;
                             });
                     };
+                    // FIXME: modal:open doesn't fire in Angular 1.5
```
dkavanagh: Is there a ticket for this?

eucaconsole/static/sass/pages/bucket_details.scss
```diff
@@ -23,21 +23,25 @@
     }
 }
 
-// CORS Configuration modal dialog
-.cors-config-modal {
+// CORS Configuration and Bucket Policy modal dialogs
+.cors-config-modal, #bucket-policy-modal {
```
dkavanagh: Curious why use class on one and id on the other?

eucaconsole/static/js/pages/bucket_details.js
kamalgill: This is similar to the Angular 1.5 issue we've seen elsewhere (e.g. CloudWatch charts dialog on resource monitoring pages).  Do you want a distinct ticket for that, or is the CW charts one adequate?

eucaconsole/static/js/pages/bucket_details.js
kamalgill: CodeMirror isn't initialized due to the Angular 1.5 modal:open issue.

eucaconsole/static/js/pages/bucket_details.js
dkavanagh: What I saw in the debugger was that modal:open was being fired, but the listener wasn't set up in time. Have you confirmed that ordering issue?

eucaconsole/static/js/pages/bucket_details.js
dkavanagh: I hate to have a band-aid for another issue that needs to be solved, right?

eucaconsole/static/js/pages/bucket_details.js
kamalgill: Yes, and it seems to be an issue (as it is everywhere else) with our custom modal.  I haven't found a fix yet -- any luck with the CloudWatch modal issue?

eucaconsole/static/js/pages/bucket_details.js
kamalgill: This is a temporary fix until we fix our custom modal for Angular 1.5.

eucaconsole/static/js/pages/bucket_details.js
kamalgill: I'm considering converting the CORS config modal to a Foundation modal... thoughts?

eucaconsole/static/js/pages/bucket_details.js
```diff
@@ -190,6 +188,29 @@ angular.module('BucketDetailsPage',
...
+                            styleActiveLine: true,
+                            autoCloseTags: true,
+                            lineNumbers: true
+                        });
+                    }
+                    // Reset to sample CORS config when re-adding post-deletion
+                    if (!$scope.hasCorsConfig && !!$rootScope.codeEditor) {
+                         $rootScope.codeEditor.setValue($scope.sampleCorsConfig);
+                    }
+                }, 300);
```
dkavanagh: Did you try with no timeout? What that does is ensure this code runs after the current $apply(). Having a timeout makes me think "race condition" and that can be unreliable. An order change by simply making this run after the current $apply() is more sound a solution.

dkavanagh: Looks like it works, I tried it. I did see that the dialog doesn't go away after saving new cors config. new ticket?

kamalgill: We need the $timeout (with a timeout value) for CodeMirror to work.  Removing that timeout breaks CodeMirror and causes the dialog to not go away after saving the CORS config.  The dialog works fine with the code as-is AFAICT.

PR #2189, opened 2017-03-15
GUI-2737 implemented programatic date range generation for select, use selected date for report

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2737

### DIFF
--- a/eucaconsole/static/js/pages/reporting_reports.js
+++ b/eucaconsole/static/js/pages/reporting_reports.js
@@ -4,6 +4,7 @@ angular.module('ReportingPage')
     var vm = this;
     vm.monthChoices = [];
     var month = new Date();
+    // set earliest year/month to show in options list (remember Jan == 0)
     while (month.getFullYear() > 2017 || month.getMonth() >= 1) {
         vm.monthChoices.push(month);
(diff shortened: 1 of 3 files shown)


eucaconsole/static/js/pages/reporting_reports.js
```diff
@@ -2,30 +2,42 @@ angular.module('ReportingPage')
 .controller('ReportsController', ['$scope', 'ReportingService', 'eucaHandleError', 'ModalService',
     function ($scope, ReportingService, eucaHandleError, ModalService) {
     var vm = this;
-    vm.monthChoices = ['January 2017', 'December 2016', 'November 2016'];
+    vm.monthChoices = [];
+    var month = new Date();
+    // set earliest year/month to show in options list (remember Jan == 0)
+    while (month.getFullYear() > 2017 || month.getMonth() >= 1) {
```
kamalgill: @dkavanagh: We'll need to update this as we get closer to the release, right?

dkavanagh: Yes, for sure. Needed to keep a start date that let us test. Not a huge downside if we don't update, but clearly ought to push this forward later.

PR #2196, opened 2017-04-07
Gui 2895 migrate CSS simplification commits from master

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2895


dkavanagh:
One concern I have (and we'll have to watch out for later) is that when cherry-picking, some files that were part of the cherry pick were not in maint-4.4. We'd need to watch for those files in any merge from maint-4.4->master (reporting, vpc, etc.. styles files)

PR #2197, opened 2017-04-10
GUI-2849 support open port range for VPC clouds

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2849


eucaconsole/static/js/widgets/listener-editor/listener-editor.js
```diff
@@ -4,7 +4,8 @@ angular.module('ELBListenerEditorModule', ['ModalModule'])
         restrict: 'E',
         scope: {
             listeners: '=ngModel',
-            cloudType: '@'
+            cloudType: '@',
+            vpcSupported: '@'
```
kamalgill: @dkavanagh This line should be `vpcEnabled: '@'`, right?

eucaconsole/static/js/widgets/listener-editor/listener-editor.js
```diff
@@ -13,6 +14,11 @@ angular.module('ELBListenerEditorModule', ['ModalModule'])
             var validPorts = [25, 80, 443, 465, 587],
                 validPortMin = 1024,
                 validPortMax = 65535;
+            if ($scope.vpcSupported) {
```
kamalgill: @dkavanagh  The VPC enabled value is passed as a string (e.g. "True" when VPC is enabled, so we should explicitly check for that to avoid false positives.

PR #2205, opened 2017-04-19
GUI-2896: Make tag editor styles consistent with other tag-like styles

kamalgill:
Minor style fix for 4.4.1 rebranding

https://eucalyptus.atlassian.net/browse/GUI-2896


dkavanagh:
I think what'd done here is good. I commented in the ticket about broader inconsistencies that I wonder are along these lines.

PR #2208, opened 2017-04-20
Gui 2727 inital reporting dash table

dkavanagh:
Not complete, but some initial data until we get EUCA-13326

https://eucalyptus.atlassian.net/browse/GUI-2727


eucaconsole/static/js/services/reporting-service.js
```diff
@@ -67,5 +67,36 @@ angular.module('ReportingServiceModule', [])
...
+                url: url
+            }).then( function (response) {
+                return response.data || {
+                    enabled: false
+                };
+            });
+        },
+
+        getInstanceUsage: function (granularity, timePeriod, fromDate, toDate, groupBy) {
+            params = {
```
kamalgill: @dkavanagh Missing a 'var' here

dkavanagh: this feels familiar!

PR #2209, opened 2017-04-21
Apr21 maint44 master

dkavanagh:
brought license header changes from maint-4.4 and updated other files in master that weren't in maint-4.4


eucaconsole/static/css/widgets/instance_selector.css
```diff
@@ -3,7 +3,7 @@
 ----------------------------------------- */
 .status.running { background-color: green; }
 
-.status.pending { background-color: #00cd00; }
+.status.pending { background-color: #00cc00; }
```
dkavanagh: I'm not sure why this is different. Where there were css conflicts, I removed the .css file and re-ran "make watch"

PR #2217, opened 2017-05-02
GUI-2910: Pass CSRF token in request body

kamalgill:
https://eucalyptus.atlassian.net/browse/GUI-2910


dkavanagh:
This is handy, but any reason not to include in merge-up that happens on Fridays?

kamalgill:
@dkavanagh: This fix is distinct from the one for GUI-2894

dkavanagh:
I see. same fix, different call.

PR #2221, opened 2017-05-09
GUI-2891: Network ACL details page WIP

kamalgill:
https://eucalyptus.atlassian.net/browse/GUI-2891

### DIFF
--- a/eucaconsole/static/js/pages/network_acl_view.js
+++ b/eucaconsole/static/js/pages/network_acl_view.js
@@ -1,0 +1,13 @@
+/**
+ * Copyright 2017 Ent. Services Development Corporation LP
+ *
+ * @fileOverview Network ACL Details Page JS
+ * @requires AngularJS
+ *
+ */
+
+angular.module('NetworkACLDetailsPage', ['TagEditorModule'])
+    .controller('NetworkACLDetailsPageCtrl', function ($scope) {
+        // pass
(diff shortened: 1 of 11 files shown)


eucaconsole/static/js/pages/network_acl_view.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright 2017 Ent. Services Development Corporation LP
+ *
+ * @fileOverview Network ACL Details Page JS
+ * @requires AngularJS
+ *
+ */
+
+angular.module('NetworkACLDetailsPage', ['TagEditorModule'])
+    .controller('NetworkACLDetailsPageCtrl', function ($scope) {
```
dkavanagh: Could simply define the module and not the controller if the intent is to define the dependency. I've done that elsewhere.

dkavanagh:
retest this please

eucaconsole/static/js/pages/network_acl_view.js
kamalgill: There will probably be code here as the NACL details page is more fully implemented.  If not, I'll remove the controller part.

dkavanagh: I was figuring the NACL could be handled like the tag editor, but guess see how that plays out.

PR #2223, opened 2017-05-10
GUI-2375 use GetRegions call for AWS logins

dkavanagh:
https://eucalyptus.atlassian.net/browse/GUI-2375


eucaconsole/layout.py
```diff
@@ -74,40 +73,38 @@ def __init__(self, context, request):
         self.default_region = ''
         self.is_vpc_supported = BaseView.is_vpc_supported(request)
         self.get_display_name = TaggedItemView.get_display_name
+        secret_key = self.request.session.get('secret_key')
+        session_token = self.request.session.get('session_token')
         if self.cloud_type == 'aws':
             self.default_region = request.registry.settings.get('aws.default.region', 'us-east-1')
             self.is_vpc_supported = True
-            self.regions = list(AWS_REGIONS)
-            if asbool(request.registry.settings.get('aws.govcloud.enabled', 'false')):
```
kamalgill: @dkavanagh: We need to keep the GovCloud region, assuming it's configured.

eucaconsole/constants/__init__.py
```diff
@@ -30,69 +30,6 @@
 """
 
 
-# AWS Regions list
-# Endpoints are handled via boto's endpoints.json
-# Note: A future release of the Eucalyptus console will pull regions via the DescribeRegions API call
-AWS_REGIONS = [
```
kamalgill: We should replace this with a mapping of region id to region label (friendly name) and look up the region label based on the id from the DescribeRegions API call.

dkavanagh: I talked with Ken about this in our 1:1. Because AWS is a secondary use case, we thought it would be OK to go with the DNS region name.

eucaconsole/layout.py
dkavanagh: Does the gov cloud not appear in get regions when the user is authorized?

dkavanagh: Hmm, can't find any evidence of that. May need to add that code back.

dkavanagh: added back below.

eucaconsole/views/policies.py
```diff
@@ -136,12 +137,16 @@ def get_controller_options_json(self):
             'cloudType': self.cloud_type,
             'actionsList': self.get_all_actions(),
             'languageCode': self.get_language_code(),
-            'awsRegions': AWS_REGIONS,
+            'awsRegions': self.get_region_choices(),
```
dkavanagh: Note: the behavior here is different now. It shows regions for current cloud. Previously, policy editor would allow location constraint to aws region when policy was created on eucalyptus (which I think was in error).

eucaconsole/constants/__init__.py
kamalgill: I think removing the friendly region names is a significant regression.  Perhaps we could use DescribeRegions to pull in *new* regions that aren't part of the static regions list, similar to how the GovCloud region is appended to the list (when configured)?

eucaconsole/layout.py
```diff
@@ -74,40 +74,43 @@ def __init__(self, context, request):
...
+                            self.default_region = region['name']
+            except BotoServerError:
+                self.has_regions = False
+            except socket.error:
+                self.has_regions = False
+            if self.cloud_type == 'aws':
+                for region in self.regions:
+                    region['label'] = AWS_REGIONS.get(region['name'].replace('-', '_'), region['name'])
+                if asbool(request.registry.settings.get('aws.govcloud.enabled', 'false')):
+                    self.regions.append(dict(name='us-gov-west-1', label='US GovCloud'))
```
kamalgill: We're appending the GovCloud region using the standard hyphen notation, but the other region keys use underscores (e.g. us_east_1).  We should probably use the standard hyphen notation everywhere to be consistent.

PR #2227, opened 2017-05-12
GUI-2890, GUI-2916: Implement Create/Delete Network ACL and allow modification of subnet-NACL associations

kamalgill:
https://eucalyptus.atlassian.net/browse/GUI-2890
https://eucalyptus.atlassian.net/browse/GUI-2916

### DIFF
--- a/eucaconsole/views/vpcs.py
+++ b/eucaconsole/views/vpcs.py
@@ -44,7 +44,7 @@ from ..forms.vpcs import (
     RouteTableForm, VPCDeleteForm, SubnetForm, SubnetDeleteForm, RouteTableSetMainForm, RouteTableDeleteForm,
     InternetGatewayForm, InternetGatewayDeleteForm, InternetGatewayDetachForm, CreateRouteTableForm,
     CreateNatGatewayForm, NatGatewayDeleteForm, NetworkACLForm, CreateNetworkACLForm, NetworkACLDeleteForm,
-    INTERNET_GATEWAY_HELP_TEXT
+    SubnetAssociateNetworkACLForm, INTERNET_GATEWAY_HELP_TEXT
 )
 from ..i18n import _
 from ..models import Notification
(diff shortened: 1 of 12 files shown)


eucaconsole/views/vpcs.py
```diff
@@ -702,21 +709,41 @@ def subnet_allocate_eips(self):
...
             return HTTPFound(location=location)
         else:
             self.request.error_messages = self.allocate_eips_form.get_errors_list()
         return self.render_dict
 
+    @view_config(route_name='subnet_associate_network_acl', renderer=VIEW_TEMPLATE, request_method='POST')
+    def subnet_associate_network_acl(self):
+        if self.subnet and self.associate_network_acl_form.validate():
+            location = self.request.route_path('subnet_view', vpc_id=self.vpc.id, id=self.subnet.id)
+            network_acl = self.request.params.get('network_acl') or None
```
dkavanagh: Trying to understand logic below. Does this mean that if a network_acl is provided that this method associates it, but won't disassociate the existing subnet_network_acl unless there isn't a network_acl provided?

kamalgill: If a network ACL is selected in the drop-down, the subnet will be associated with that, which is covered by the block with the check at line 730.  

If no network ACL is selected (block with condition at line 736), that implies the subnet should be disassociated from an existing network ACL, assuming an explicit association exists.  If no implicit association exists, the subnet will use the VPC's default network ACL and will continue to do so.

If this isn't clear, I can add these notes as comments in the method's docstring.

eucaconsole/views/vpcs.py
```diff
@@ -702,21 +709,41 @@ def subnet_allocate_eips(self):
...
+            if network_acl:
+                self.log_request(_(u"Associating subnet with network ACL {0}").format(network_acl))
+                msg = 'Successfully associated subnet with network ACL {0}'.format(network_acl)
+                with boto_error_handler(self.request, location):
+                    self.vpc_conn.associate_network_acl(network_acl_id=network_acl, subnet_id=self.subnet.id)
```
dkavanagh: I see. this call uses the "ReplaceNetworkAclAssociation" API call, and so disassociates the former one. got it.

